PR #1, opened 2018-08-03
WIP: Move base_rest from https://github.com/akretion/odoo-shopinvader

lmignon:


base_rest/components/service.py
```diff
@@ -0,0 +1,135 @@
...
+    # to the api, the werkzeug request interpret params as unicode
+    # so we have to convert params from string to float to int
+    if val:
+        return int(float(val))
+    else:
+        return None
+
+
+def to_bool(val):
+    return val in ('true', 'True', '1', True)
```
kryskool, line 38: Do we add **Y** value for system use char(1) as Boolean value ?
https://www.devart.com/dotconnect/oracle/docs/CharAsBoolean.html

lmignon: @kryskool Thank you for your comment. This module is a WIP. But I'll take into account your comment. The API is not yet finalized see https://github.com/OCA/rest-framework/issues/2

kryskool: @lmignon thanks for the feedback
i take a look at #2 
Regards,

PR #3, opened 2018-08-09
[10.0] base_rest: Move from https://github.com/akretion/odoo-shopinvader

lmignon:
suppersedes #1 (wrong branch)
superseded by #5

PR #5, opened 2018-08-10
[10.0] base_rest: Add swagger frontend and generate OpenAPI specs for sevices 

lmignon:
based  on #3 
moved from https://github.com/akretion/odoo-shopinvader/pull/144
* [X] Add a swagger interface to test the services
* [X] Generate openapi definition for each services
image
image

* [X] Backward compatibility? (Response schema is required)
* [X] Missing tests
* [X] Code cleaning
* [X] Binary format for request/response  (ie post or retrieve a file)


base_rest/components/service.py
```diff
@@ -151,8 +150,7 @@ def _secure_output(self, method, response):
         v = Validator(schema, purge_unknown=True)
         if v.validate(response):
             return v.document
-        _logger.error("Invalid response %s", v.errors)
-        raise UserError(_('Invalid Response'))
+        raise UserError(_('Invalid Response %s') % v.errors)
```
sbidoul: IMO output validation error is a system error, not a user error

sbidoul: because it's the developer's fault, not the user's

lmignon: it makes sense. I'll change

sbidoul:
I think most of the TODO's in the PR description are done right?

sbidoul (approved):
This has been used successfully in several projects.

Thanks for this great framework.

lmignon:
@sbidoul Yes they are done.

sebastienbeau (approved):
Thansk for this awsome work, used in production

pedrobaeza:
This is not in v11 yet, right?

lmignon:
@pedrobaeza v10 and v11 provides the same features and share the same code. IOW Both versions are aligned

pedrobaeza:
Thanks for the confirmation!

PR #7, opened 2018-09-01
[11.0] [MIG] base_rest, base_rest_demo: Migration to 11.0

lmignon:


base_rest/http.py
```diff
@@ -43,11 +43,18 @@ def wrapJsonException(exception):
     get_original_headers = exception.get_headers
 
     def get_body(environ=None):
-        return JSONEncoder().encode({
+        res = {
             'code': exception.code,
             'name': escape(exception.name),
             'description': exception.get_description(environ)
+            }
+        if _logger.isEnabledFor(logging.DEBUG):
```
sbidoul: This sounds dangerous. Enabling `--log-level=debug` globally will have the (surprising IMO) side effect of enabling stack traces in base_rest (which is often considered a security issue).

I'd prefer an option in the config file, that we could then manage with `server_environment`.

lmignon: @sbidoul Thank you for this insightful comment. I'll change the code to use an option into the config file.

lmignon: @sbidoul What do you think about https://github.com/OCA/rest-framework/pull/5/commits/317dcb4a91ae049af6c6778c6e6ef54222f31f77

sbidoul: I made a couple of comments on the commit.

.travis.yml
```diff
@@ -3,7 +3,7 @@ sudo: false
 cache: pip
 
 python:
-  - "2.7.13"
+  - "3.6"
```
sbidoul: For 11.0 we need to test on 3.5 which is the minimum version supported by Odoo.

setup/base_rest_demo/odoo/addons/__init__.py
```diff
@@ -0,0 +1 @@
+__import__('pkg_resources').declare_namespace(__name__)
```
sbidoul: no init.py for 11

sbidoul (approved):
A couple of minor changes, otherwise +1

lmignon:
Thank you @sbidoul  Changes done

sebastienbeau:
Ok for me :+1:

sebastienbeau (approved):

PR #11, opened 2018-10-17
[11.0][FIX] overwrite the dependencies in the setup for base_rest 'accept_l…

RobinetDenisAcsone:
…anguage'-> 'parse-accept-language'


RobinetDenisAcsone:
@lmignon

lmignon (approved):
Thank you @RobinetDenisAcsone

lmignon:
fast tracking since the change is trivial and related to setup.py (the code is right in 10.0)

PR #13, opened 2018-10-30
[11.0][ADD] Documentation to authenticate as Odoo user via JSON-RPC

### LINKED ISSUE #12: Support for authentication
Unless I overshot snippets of code in base_rest, it looks like the module does not allow to authenticate to private services in a RESTful way. I guess most of the APIs written with this framework will only be available to authenticated users, so it would be great to add support for authentication with a standard request shared among services.

By the way, how do you manage authentication with the current module? An RPC call to login and get the session?

warp10:
Closes #12


base_rest/readme/USAGE.rst
```diff
@@ -114,4 +114,31 @@ to access to your services
...
 The HTTP GET 'http://my_odoo/my_services_api/ping' will be dispatched to the
-method ``PingService.search``
\ No newline at end of file
+method ``PingService.search``
+
+Authentication
+~~~~~~~~~~~~~~
+
+The authentication to use services must be specified into the controller by the
+key ``_default_auth``. Then you can use any authentication mechanism supported
```
lmignon:
```suggestion
key ``_default_auth``. 


.. code-block:: python

    from odoo.addons.base_rest.controllers import main

    class MyRestController(main.RestController):
        _root_path = '/rest_api/'
        _collection_name = my_module.services
        _default_auth  = 'user'

Then you can use any authentication mechanism supported
```

lmignon (changes requested):
@warp10 Thank you for your proposal. I propose to add a code block to illustrate where to add the `_default_auth` key.

dreispt (approved):

dreispt:
@lmignon This is an important piece of documentation.
Can it be merged and forward ported?

lmignon:
@dreispt I've to fix the travis....

PR #14, opened 2018-10-31
[WIP] Alternative base rest module

LeartS:
Hi guys,
This is a module to implement rest controllers that I've worked on for a project .
Unfortunately when I started working on it I wasn't aware of the work being done on the `base_rest` module. The need it solves is pretty much the same. From what I saw of the `base_rest` module, this one does less stuff (e.g. it doesn't consider endpoint documentation or definitions) but it's easier to define simple rest controllers (no service - controller structure). Also I implemented a `rest.mixin` abstract model that I think it's very cool to quickly expose recordsets as json objects, give it a look.

I wouldn't consider this module "complete", but it works and it's already usable. We use it to expose REST API for a iOS app that must be able to see and edit partners, crm leads, activities and some other custom models.

The readme is not complete but has most of the important details. I'll try to add an example module.
I'm not sure how we want to proceed and if this is useful. We could have the two alternative modules so that users can choose which one to use according to their needs / preferences, merge one into the other, or just take some ideas of this one (e.g. `rest.mixin`) and put it into `base_rest`, or just disregard this module completely if you feel it doesn't have any advantage over `base_rest`. I just wanted to share the work, I think there are some useful pieces in here.

(p.s. I used git subtree to merge the module history, which was tracked in a single module repo, into this one)


lmignon:
@LeartS Thank you for sharing your code. Indeed, both approach are at some point different but also have some common parts. 
The main idea into base_rest is to provide a framework to build rest API on top of odoo with a strong input/output data validation. The intent is not to provide a rest API to expose directly Odoo models since IMO it's important to avoid a strong coupling between an external API and odoo internals. That being said we could also provide a way to implement such features while warning of the limitations of such an approach..
Your code comes with a lot of good ideas. IMO we should find a way to make base_rest and base_request_rest converge. (into one addon or into different addons with the same base or ....). I don't have time at the moment to analyze mode deeply your module in order to propose a path to integrate your work into oca/rest-framework. I'll try to take this time in the next weeks... (any ideas are welcome)
Once again, thank you for your proposal.

PR #17, opened 2018-11-13
[10.0][FIX] Fix little issue if no '__doc__' filled

acsonefho:
**Issue**
If no `__doc__` filled on the target method, an exception is raised

**Fix**
Just add an empty text when no `__doc__` on the function.


lmignon (approved):
Good catch @acsonefho

PierrickBrun (approved):

PR #31, opened 2019-06-21
[10.0][IMP][FIX] Validator in case of inherit

acsonefho:
**Current behaviour:**
A Component with a public method (without leading `_`) without validator should be considerate as private.
Because in case of inherit of Components with public methods, they'll be exposed/visible to the service. And as they don't have any validator, they shouldn't be displayed (swagger).

**Why a fix?**
Because in case of inherit (with public method without validators), the swagger interface is broken.

**Example**
```
class ComponentA(Component):
    _name = "comp.a"

    def get_uncrypted_pwd_current_user():
        return self.env.user.password

class ComponentB(Component):
    _name = "comp.b"
    _inherit = ["comp.a"]

    def get_user_name(id):
        return self.env['res.users'].browse(id).name

    def _validator_get_user_name():
        return ...
```
The `get_uncrypted_pwd_current_user()` is visible into swagger.

### DIFF
--- a/base_rest/components/service.py
+++ b/base_rest/components/service.py
@@ -297,6 +297,12 @@ class BaseRestService(AbstractComponent):
                     }
                 })
             input_schema = self._get_input_schema(name)
+            # If we don't have an input schema (None), we have to considerate
+            # the function as private.
+            # Useful in case of inheritance of Components with
+            # functions without leading "_".
+            if not isinstance(input_schema, dict):
+                continue
             output_schema = self._get_output_schema(name)
             json_input_schema = cerberus_to_json(
                 input_schema)


sbidoul:
@acsonefho I'm not familiar enough with base_rest to judge if it makes sense to hide a method that has no schema. However inheritance is an orthogonal concept, so I don't really understand why the two are being discussed in the same PR.

It sounds logical that public methods remain public in inherited components, so if that is an issue in a specific case, either the methods of the parent component must be prefixed with `_` or the extension must be done with composition instead of inheritance.

BTW, is there a real public method somewhere named `get_uncrypted_pwd_current_user`? Or is it just and example?

acsonefho:
> @acsonefho I'm not familiar enough with base_rest module to judge if it makes sense to hide a method that has no schema. However inheritance is an orthogonal concept, so I don't really understand why the two are being discussed in the same PR.

It's because a Component (coming from payment_gateway repo) has some public methods (`generate`, `capture` etc). So into the swagger interface, the `base_rest` module try to load the validator for these methods (as they don't exist, it's return `None`). Later (into `cerberus_to_json`) we're loading the parameter as a dict. As it's `None`, it failed.
IMO, methods `generate`, `capture` etc shouldn't be visible into swagger because they shouldn't be called by the front (but maybe I'm wrong?).
So I had 2 options:
- Return an empty `dict` into the validator loader (but in this case, these methods become visible into swagger);
- Ignore the function if no validator (this MR).
- Or maybe these methods (`generate`, `capture` etc) should be private (with leading `_`)?
> 
> It sounds logical that public methods remain public in inherited components, so if that is an issue in a specific case, either the methods of the parent component must be prefixed with `_` or the extension must be done with composition instead of inheritance.
> 
> BTW, is there a real public method somewhere named `get_uncrypted_pwd_current_user`? Or is it just and example?

Of course it's just an example :)

lmignon:
@acsonefho IMHO  in python, inheritable private methods must be prefixed by ' _' . It's not the intend of base_rest to redefine an other way to define private methods or introduce exceptions to this orthogonal concept.
The problem is not into base_rest but in the way your services are developed.

lmignon:
@acsonefho IMO, thinking further, I think we should put in place a mechanism that detects this type of inconsistency (public method without a validator) and prevents Odoo from starting in such cases. (or at least one that displays a specific error...)
The fix to write for your case is probably to modify the code into payment_gateway ..

sbidoul:
> I think we should put in place a mechanism that detects this type of inconsistency (public method without a validator) and prevents Odoo from starting in such cases. (or at least one that displays a specific error...)

Perhaps create an issue here to track this improvement.

But indeed, an error sounds better than hiding the method.

lmignon:
@acsonefho @sbidoul I've created an issue to track this improvement #32

PR #37, opened 2019-08-01
[IMP] let _validator methods return cerberus schemas

sbidoul:


lmignon:
@sbidoul This change improve flexibility. For your use case, do you have tried by defining your schema as 
```python
schema = {
  'name': {'type': 'string'},
  'allow_unknown': True,
  ....
}
```
see https://docs.python-cerberus.org/en/stable/usage.html#allowing-the-unknown

sbidoul:
@lmignon I don't think it would work. IIUC one can use `allow_unknown` in a subschema, but not at that level where only field names can be used as keys.

lmignon:
> @lmignon I don't think it would work. IIUC one can use `allow_unknown` in a subschema, but not at that level where only field names can be used as keys.

You're right I misread the documentation.

lmignon (approved):

PR #41, opened 2019-08-22
[12.0][ADD] datamodel: New addon allowing you to define dataclass like mode…

lmignon:
…ls serialisable/deserializable to/from json/yaml

Part of the code is inspired from [component](https://github.com/OCA/connector/tree/12.0/component)
This addons is based on [marshmallow](https://marshmallow.readthedocs.io/en/stable/) and [marshmallow-objects](https://github.com/SVilgelm/marshmallow-objectsl)

The idea behind this new addon is to provide a more pythonic way to define and manipulate data exchanged by REST services


lmignon:
ping @sbidoul @simahawk @sebastienbeau  you might be interested. The code into the tests illustrate the idea and how it works https://github.com/OCA/rest-framework/pull/41/files#diff-7537e6fc7a03a7bc1064cfe01cc785a5R313

datamodel/builder.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+
+Datamodels Builder
+==================
+
+Build the datamodels at the build of a registry.
+
+"""
+import odoo
+from odoo import api, models
```
apineux:
```suggestion
from odoo import api, models, modules
```
You can therefore remove import odoo

datamodel/builder.py
```diff
@@ -0,0 +1,107 @@
...
+        return datamodels_registry
+
+    def build_registry(self, datamodels_registry, states=None,
+                       exclude_addons=None):
+        if not states:
+            states = ('installed', 'to upgrade')
+        # lookup all the installed (or about to be) addons and generate
+        # the graph, so we can load the datamodels following the order
+        # of the addons' dependencies
+        graph = odoo.modules.graph.Graph()
```
apineux:
```suggestion
        graph = modules.graph.Graph()
```
Link to my previous suggestion

datamodel/builder.py
```diff
@@ -0,0 +1,107 @@
...
+    _name = 'datamodel.builder'
+    _description = 'Datamodel Builder'
+
+    _datamodels_registry_cache_size = DEFAULT_CACHE_SIZE
+
+    @api.model_cr
+    def _register_hook(self):
+        # This method is called by Odoo when the registry is built,
+        # so in case the registry is rebuilt (cache invalidation, ...),
+        # we have to to rebuild the datamodels. We use a new
```
apineux:
```suggestion
        # we have to rebuild the datamodels. We use a new
```
tipo

datamodel/core.py
```diff
@@ -0,0 +1,373 @@
...
+# unfortunately can't use because it's an instance method and should have been
+# a @staticmethod
+def _get_addon_name(full_name):
+    # The (Odoo) module name can be in the ``odoo.addons`` namespace
+    # or not. For instance, module ``sale`` can be imported as
+    # ``odoo.addons.sale`` (the right way) or ``sale`` (for backward
+    # compatibility).
+    module_parts = full_name.split(".")
+    if len(module_parts) > 2 and module_parts[:2] == ["odoo", "addons"]:
+        addon_name = full_name.split(".")[2]
```
apineux:
```suggestion
        addon_name = module_parts[2]
```

datamodel/core.py
```diff
@@ -0,0 +1,373 @@
...
+    else:
+        addon_name = full_name.split(".")[0]
```
apineux:
```suggestion
        addon_name = module_parts[0]
```

datamodel/core.py
```diff
@@ -0,0 +1,373 @@
...
+        if isinstance(parents, str):
+            parents = [parents]
+        elif parents is None:
+            parents = []
+
+        if cls._name in registry and not parents:
+            raise TypeError(
+                "Datamodel %r (in class %r) already exists. "
+                "Consider using _inherit instead of _name "
+                "or using a different _name." % (cls._name, cls)
```
apineux, line 282: String formating with % has been de-emphasized in Python 3 so it may be a good Idea to change it with .format()

datamodel/core.py
```diff
@@ -0,0 +1,373 @@
...
+            )
+
+        # determine the datamodel's name
+        name = cls._name or (len(parents) == 1 and parents[0])
+
+        if not name:
+            raise TypeError("Datamodel %r must have a _name" % cls)
```
apineux, line 289: Same comment as previously, it is recommended to use .format()

datamodel/core.py
```diff
@@ -0,0 +1,373 @@
...
+
+        # all datamodels except 'base' implicitly inherit from 'base'
+        if name != "base":
+            parents = list(parents) + ["base"]
```
apineux, line 293: Event if it is not important.
extend() is faster than +
In this case a append() seems to be the best

datamodel/core.py
```diff
@@ -0,0 +1,373 @@
...
+
+        # create or retrieve the datamodel's class
+        if name in parents:
+            if name not in registry:
+                raise TypeError(
+                    "Datamodel %r does not exist in registry." % name
```
apineux: .format()

datamodel/core.py
```diff
@@ -0,0 +1,373 @@
...
+                )
+
+        # determine all the classes the datamodel should inherit from
+        bases = LastOrderedSet([cls])
+        for parent in parents:
+            if parent not in registry:
+                raise TypeError(
+                    "Datamodel %r inherits from non-existing datamodel %r."
+                    % (name, parent)
```
apineux, line 306: .format()

datamodel/core.py
```diff
@@ -0,0 +1,373 @@
...
+                for base in parent_class.__bases__:
+                    bases.add(base)
+            else:
+                bases.add(parent_class)
+                parent_class._inherit_children.add(name)
+
+        if name in parents:
+            DatamodelClass = registry[name]
+            # Add the new bases to the existing model since the class into
+            # the registry is could already be used into an inherit
```
apineux:
```suggestion
            # the registry could already be used into an inherit
```

datamodel/core.py
```diff
@@ -0,0 +1,373 @@
...
+    def _complete_datamodel_build(cls):
+        """ Complete build of the new datamodel class
+
+        After the datamodel has been built from its bases, this method is
+        called, and can be used to customize the class before it can be used.
+
+        Nothing is done in the base Datamodel, but a Datamodel can inherit
+        the method to add its own behavior.
+        """
+
```
apineux, line 362:
```suggestion
pass
```
To avoid "indent expected" error

datamodel/tests/common.py
```diff
@@ -0,0 +1,224 @@
...
+        # resolve an inheritance issue (common.TransactionCase does not call
+        # super)
+        common.TransactionCase.setUp(self)
+        DatamodelMixin.setUp(self)
+
+
+class SavepointDatamodelCase(common.SavepointCase, DatamodelMixin):
+    """ A SavepointCase that loads all the datamodels
+
+    It it used like an usual Odoo's SavepointCase, but it ensures
```
apineux:
```suggestion
    It is used like an usual Odoo's SavepointCase, but it ensures
```

datamodel/tests/test_build_datamodel.py
```diff
@@ -0,0 +1,419 @@
...
+            ),
+            self.datamodel_registry["datamodel1"].__bases__,
+        )
+
+    def test_prototype_inherit_bases(self):
+        """ Check __bases__ of Datamodel with _inherit and different _name """
+
+        class Datamodel1(Datamodel):
+            _name = "datamodel1"
+
```
apineux:
```suggestion
```

datamodel/tests/test_build_datamodel.py
```diff
@@ -0,0 +1,419 @@
...
+            ),
+            self.datamodel_registry["datamodel4"].__bases__,
+        )
+
+    def test_final_class_schema(self):
+        """Check the MArshmallow schema of the final class"""
+
+        class Datamodel1(Datamodel):
+            _name = "datamodel1"
+
```
apineux:
```suggestion
```

apineux (commented):
@lmignon Very nice dev
I made small (and not mandatory) suggestions if you want.

lmignon:
Thank you @apineux

blaggacao:
Maybe you might be interested in new python native `@dataclass` as a specification variant, being typed, they are slick as hell.

Not sure about the specific % of overlap with the purpose of data models, though. 

Just smells like...

blaggacao:
Ha! https://pypi.org/project/dataclasses-serialization/

blaggacao:
Btw if there was a pythonic inheritance for better static code analysis support, with a string indirection support (similar to what Django does), this would be a nice place to experiement ;-)

For this to work, teh data class in the MRO must hold an in memory reference to the database where it's intitialized, and a resolver must point to the appropriate (non)-terminal in the database context.

This essentially simplifies the registry as a welcome side effect, and ultimately cache can be implemented on the dataclass, too.

blaggacao:
```python
@dataclass
class Partner(odoo.addons.partner.dataclasses.Partner):
    class Meta:
        _dbs = []  # initialized on those dbs

    name: str = "Peter"  # or a default factory


Partner.register()  # with the resolver that inhabits the metaclass...
```

This would "anchor" models to their first definition in plain python - as opposed to accessing through a relatively opaque (string type!) hash from the environment.

Class names become semantically significant. (DRY)

... was playing with this thought recently and how it would make Odoo a more pleasant framework.

Here might be a place to test.

lmignon:
@blaggacao I'm aware of `@datacalss` I've also investigated some times and made some experiment to use a more pythonic way to implement the functionalities provided by this addon (data validation / serialization / deserialization / auto description through json schema, ...) .
Some of these experiments was based on [pythdantic](https://pypi.org/project/pydantic/)
Odoo as frameworks comes with some constrains. The most challenging one is that the same code / runtime can be used to server different databases where the list of installed addons can differ. That means that it's not possible to get RID of the registry to get the right model defitinion for the current database according to the addons installed into the database.

blaggacao:
I also "hate" the registry :-) I've reasoned for a way to work around it and it should be theoretically possible by overloading function and attribute resolvers to make them db aware instead of building custom models. This should work as long as no use is made of the fields class call overrides such as `selection_add` but that's not within the scope of dataclasses, I guess...

BTW I'm working on doDoo to make Odoo as a server more well-behaved. I'm planning to add a `doDoo run graphql` as a standalone (and individually scalable) service to run graphql endpoints.

How does this proposal for dataclasses relate to graphene's type system?

Correct me if I'm wrong, but maybe it would be true leap frogging to built data classes around graphql (which is awesome!) and rather built implementations for other serializations (than json) or compatibility shims behaving as REST endpoints.

https://github.com/xoe-labs/dodoo#readme

I used data classes for a configs shim, works pretty good, as they are somewhat immutable as a nice feature.

blaggacao:
I'm going for strawberry graphql. It's awesome (with dataclasses)! Probably exactly what's needed... Have to check how to handle starlett async wsgi ASGI in an Odoo context. I'll push draft code in a while so you can have a look.

It only relies on Odoo for auth and env, the rest is a somewhat more modern server tooling...

For the db problem, there is an easy solution, during authentication, we can enumerate the installed modules for the database and then put them as decorators on fields.

Esentially hooking on an credential system for defining acccesss scopes infered from the database.

We can then suppress the types that are only supported with "higher level" databases (more modules installed)

blaggacao:
If the goal is to put schema files on every module, we might as well just load the graphql server differently searching the `odoo.addons` namespace for schema attributes based on the list in `odoo.modules.module.loaded`

No complicated stuff, but a three liner.

kevinkhao:
I think it is worth mentioning somewhere you need marshmallow_objects version 1.0.23 for it, 2.0.0 onwards does not work. Newest marshmallow version also fails one test on this. Works with 3.2.2

lmignon:
@kevinkhao Some changes are into #48 The  required version is specified there https://github.com/OCA/rest-framework/pull/48/files#diff-b4ef698db8ca845e5845c4618278f29aR8

Unfortunately all this work stay pending since there is no review (nor interest?)

kevinkhao:
I see, well it is a cool piece of code and we would like to use it. Personally one barrier to review it is that it is outside my technical comfort zone, but I will try to contribute. Will talk about it with @sebastienbeau

lmignon:
@kevinkhao Functional tests are also very valuable as review! :smirk:

kevinkhao:
Currently, validators allow only for operations on static data.

```
class SaleOrderDatamodel(BaseDatamodel):
    _name = "sale.order"

    def _validate_my_obj(self):
        if not self.get('some_data'):
            return False

    customer = fields.NestedModel(
        "my.obj",
        required=True,
        validate=_validate_my_obj,
    )
```

Do you think it would be possible to do something like this, or is that too out of scope for Datamodel itself ? 

```
class SaleOrderDatamodel(BaseDatamodel):
    _name = "sale.order"

    def _validate_my_obj(self, myObj):
        partner_ids = self.env["res.partner"].search(
            [("some_res_partner_field", "=", myObj['some_data'])]
        )
        if not partner_ids:
            return False

    customer = fields.NestedModel(
        "my.obj",
        required=True,
        validate=_validate_my_obj,
    )
```

lmignon:
@kevinkhao du you have tried https://marshmallow.readthedocs.io/en/stable/quickstart.html#field-validators-as-methods

kevinkhao:
@lmignon inside such a method decorated by @validates, it acts on the Schema-level, so self evaluates to a Schema and there is no access to Odoo's env

lmignon:
@kevinkhao As the datamodel registry is available into the schema context, it should be possible to also have the env available into the same context or into the registry.... seems not so difficult

kevinkhao:
@lmignon not so difficult for you, I believe that :smile:
This was a great opportunity to refresh on Python fundamentals and Odoo internals. I made a PR https://github.com/acsone/rest-framework/pull/1 to add a convenience env to be used that way

lmignon:
@kevinkhao Implemented here https://github.com/OCA/rest-framework/pull/48/commits/7027645a905f06bcc96f5f7d3cf3d49ce03dde9b :smirk:

kevinkhao:
@lmignon thanks for the news, that is very interesting. Will report with some tests soon.

kevinkhao (approved):
Works great and as expected, very useful functionality.

kevinkhao:
We're using this module in a WIP collection of modules for json-like imports/exports: 
https://github.com/akretion/sale-import/tree/WIP

lmignon:
> Works great and as expected, very useful functionality.

Thank you for the feedback!

datamodel/builder.py
```diff
@@ -0,0 +1,97 @@
...
+
+from .core import DEFAULT_CACHE_SIZE, DatamodelRegistry, _datamodel_databases
+
+
+class DatamodelBuilder(models.AbstractModel):
+    """ Build the datamodel classes
+
+    And register them in a global registry.
+
+    Every time an Odoo registry is built, the know datamodels are cleared and
```
gurneyalex:
```suggestion
    Every time an Odoo registry is built, the known datamodels are cleared and
```

datamodel/builder.py
```diff
@@ -0,0 +1,97 @@
...
+    rebuilt as well.  The Datamodel classes are built using the same mechanism
+    than Odoo's Models: a final class is created, taking every Datamodels with
```
gurneyalex:
```suggestion
    as Odoo's Models: a final class is created, taking every Datamodels with
```

datamodel/builder.py
```diff
@@ -0,0 +1,97 @@
...
+    a ``_name`` and applying Datamodels with an ``_inherits`` upon them.
+
+    The final Datamodel classes are registered in global registry.
+
+    This class is an Odoo model, allowing us to hook the build of the
```
gurneyalex:
```suggestion
    This class is an Odoo model, allowing us to hook the building of the
```

datamodel/builder.py
```diff
@@ -0,0 +1,97 @@
...
+    datamodels at the end of the Odoo's registry loading, using
```
gurneyalex:
```suggestion
    datamodels at the end of the loading of Odoo's registry, using
```

datamodel/builder.py
```diff
@@ -0,0 +1,97 @@
...
+    ``_register_hook``. This method is called after all modules are loaded, so
```
gurneyalex:
```suggestion
    ``_register_hook``. This method is called after all modules have been loaded, so
```

datamodel/builder.py
```diff
@@ -0,0 +1,97 @@
...
+    _name = "datamodel.builder"
+    _description = "Datamodel Builder"
+
+    _datamodels_registry_cache_size = DEFAULT_CACHE_SIZE
+
+    @api.model_cr
+    def _register_hook(self):
+        # This method is called by Odoo when the registry is built,
+        # so in case the registry is rebuilt (cache invalidation, ...),
+        # we have to rebuild the datamodels. We use a new
```
gurneyalex, line 47: how expensive is this operation?

datamodel/core.py
```diff
@@ -0,0 +1,401 @@
...
+        addon_name = module_parts[0]
+    return addon_name
+
+
+class DatamodelDatabases(dict):
+    """ Holds a registry of datamodels for each database """
+
+
+class DatamodelRegistry(object):
+    """ Store all the datamodel and allow to retrieve them by name
```
gurneyalex, line 49:
```suggestion
    """ Store all the datamodels and allow to retrieve them by name
```

datamodel/core.py
```diff
@@ -0,0 +1,401 @@
...
+    This is an OrderedDict, because we want to keep the registration order of
+    the datamodels, addons loaded first have their datamodels found first.
+
+    The :attr:`ready` attribute must be set to ``True`` when all the datamodels
+    are loaded.
+
+    """
+
+    def __init__(self, cachesize=DEFAULT_CACHE_SIZE):
+        self._datamodels = OrderedDict()
```
gurneyalex, line 62: why not inherit from OrderedDict?

datamodel/core.py
```diff
@@ -0,0 +1,401 @@
...
+
+    This is an OrderedDict, because we want to keep the registration order of
+    the datamodels, addons loaded first have their datamodels found first.
+
+    The :attr:`ready` attribute must be set to ``True`` when all the datamodels
+    are loaded.
+
+    """
+
+    def __init__(self, cachesize=DEFAULT_CACHE_SIZE):
```
gurneyalex, line 61: cachesize is not used.

datamodel/core.py
```diff
@@ -0,0 +1,401 @@
...
+
+    def load_datamodels(self, module):
+        if module in self._loaded_modules:
+            return
+        for datamodel_class in MetaDatamodel._modules_datamodels[module]:
+            datamodel_class._build_datamodel(self)
+        self._loaded_modules.add(module)
+
+
+# We will store a DatamodeltRegistry per database here,
```
gurneyalex:
```suggestion
# We will store a DatamodelRegistry per database here,
```

datamodel/core.py
```diff
@@ -0,0 +1,401 @@
...
+# it will be cleared and updated when the odoo's registry is rebuilt
```
gurneyalex:
```suggestion
# it will be cleared and updated when odoo's registry is rebuilt
```

datamodel/core.py
```diff
@@ -0,0 +1,401 @@
...
+            super(MetaDatamodel, self).__init__(name, bases, attrs)
+
+            return
+
+        # If datamodels are declared in tests, exclude them from the
+        # "datamodels of the addon" list. If not, when we use the
+        # "load_datamodels" method, all the test datamodels would be loaded.
+        # This should never be an issue when running the app normally, as the
+        # Python tests should never be executed. But this is an issue when a
+        # test creates a test datamodels for the purpose of the test, then a
```
gurneyalex:
```suggestion
        # test creates a test datamodel for the purpose of the test, then a
```

datamodel/core.py
```diff
@@ -0,0 +1,401 @@
...
+        # second tests uses the "load_datamodels" to load all the addons of the
```
gurneyalex:
```suggestion
        # second test uses the "load_datamodels" to load all the addons of the
```

gurneyalex (approved):

etobella (approved):
I think that the only thing needed is to finish the readme and squash commits, isn't it @lmignon ?

etobella:
@lmignon , If you have too much work, we can take the PR and finish it :smile:

etobella:
@lmignon Do you need some help finishing this PR?

lmignon:
@etobella This one is included into #48. Merging now...

lmignon:
merged with #48

etobella:
Gr🎱 !!! I thought there were two PRs.

Thanks for your great work

PR #40, opened 2019-08-08
[12.0][UPD] Update base_rest module logo

marcelsavegnago:
- Update base_rest module logo

PR #43, opened 2019-08-28
[10.0][UPD] Update base_rest module logo

lmignon:
backport of #40


pedrobaeza:
Not sure what has happened with CLA, as the partner was there with the tag. I'll check and tell you back...

lmignon:
No stress @pedrobaeza Thank you for the support.

pedrobaeza:
OK, I see the problem: the registered email is not the same (not showing it due to GDPR privacy question, but it's a gmail one).

@marcelsavegnago should we switch the email address to the new one?

Issue #2, opened 2018-08-06
[11.0] base_rest: RFC: Improve request/response Schema description


lmignon:
## Current context

In base_Rest, information exchanged by services is validated according to a Cerberus schema defined as dict.

```python
    address_request_schema = {
        'name': {'type': 'string', 'required': True, 'empty': False},
        'street': {'type': 'string'},
    }
```

Defining such schemas could become boring and lead to a hard to read/maintain result when the schema is complex. This kind of definition is also hard to extend, reuse etc...

## Proposal

After much reflection and discussion with @sbidoul and some research into the web, I think that I'll base the schema definition of our data transfer objects on the  [pydantic](https://pydantic-docs.helpmanual.io) library. This library comes with a clean and standard way to define the structure of data transfer objects. It also provides methods to serialize/deserialize defined data transfer object to/from json and to generate the jsonschema from the objet definition.

~~Odoo already provides a clean api to define models. This API provides a pythonic way to define fields and relation between models. It also provides a clean way to override and extend models across modules. The Idea is to define a new base class based on BaseModel to define the messages exchanged by the services.~~

```python

class RestModel(models.BaseModel):
    _name='rest.model'
    
    def from_json(self, json_data):
        return self.new(....)

    def to_json(self):
        return {}      
    
    def validate(self):
        return True

    def to_json_schema(self):
        """ Return the definition of the model as json schema
        see http://json-schema.org/
        """
        retrun {}


class AddressRequest(RestModel):
    _name = 'address.request'

    name = fields.Char(required=True)
    street = fields.Char()

    ....
```

~~Services will be annoted to specify the request/response models instead of the current implicit properties. This annotation will also be used to say that the request  / response is a list or not (default)~~ see #6

### Expected benefits:

* More modular
* Easier to extend / reuse / redefine
* Easier to manipulate into the services implementation. `my_address['name'] vs my_address.name`
* Easier to document (thanks to `string` and `help`attributes of the fields we'll be able to provide a description into the json schema)
* ....

sbidoul:
:+1:

lmignon:
@sebastienbeau @sbidoul 
I wonder if it makes sense to base the definition of messages exchanged on Odoo models. At this stage I have the feeling that the Odoo models are too complex objects that come with a lot of features specific to their use by the ORM. The'id' field for example is implicit in these models. We also have a strong management of the cache relative to an ORM without speaking of the advanced functionnalities allowing the definition of calculated fields etc....
[...]

PR #48, opened 2019-10-17
[12.0] base_rest: New api

### LINKED ISSUE #6: [11.0] [?10.0?] base_rest: RFC: Annotation for the methods exposed by the services
## Current context

The current implementation is based on implicit rules:
* public methods defined into a component inheriting of `'base.rest.service'` are implicitly available into the REST API. 
```python
class PartnerService(Component):
    _inherit = 'base.rest.service'
    _name = 'partner.service'
    _usage = 'partner'
    _collection = 'base.rest.demo.private.services'
    _description = """
Partner Services
    """

    def get(self, _id):
        """
        Get partner's informations
        """
        return self._to_json(self._get(_id))
```
* schemas used to validate the data transferred by these services (in/out) are provided by protected methods (_validator_...). The link between services methods and methods providing validation schemas is based on naming conventions.

```python
    def get(self, _id):
        """
        Get partner's informations
        """
        return self._to_json(self._get(_id))

    # Validator
    def _validator_get(self):
        return {}

    def _validator_return_get(self):
        res = self._validator_create()
        res.update({
            'id': {'type': 'integer', 'required': True, 'empty': False},
        })
        return res
```

* It's possible to return a result in a different format than JSON. Nevertheless there is no way to document/enforce this type of result

```python
from odoo.http import Response
...
    def get_image(self, _id):
        ....
        return Response(
            output.getvalue(),
            headers={
                'Content-Type': mimetype,
                'Content-Disposition': 'inline; filename="%s"' % filename,
            })
```

* By default, the base_rest addon register generic http routes. To avoid the call of methods that could do updates by an HTTP GET request (which is against the HTTP) we enforce only 2 methods available by default via HTTP GET. If you need to add another method, you must declare it into the controller.

* It's not possible to specify that a method takes as input our returns as output a list of items... 

## Proposal

I I think that the development of a specific decorator to annotate the exposed methods would allow us to pass from an implicit mode to an explicit mode for the various stated elements. This would also allow us to have a stronger semantic in the definition of methods.

```python

    @restapi.method(
        input = None,
        output = File(),
        auth='User',
        method='GET'
    )
    def get_image(self, _id):
        ....
        return {
            'mimetype': ...
            'filename': ....
            'data': output.getvalue(),
        }

    @restapi.method(
        input = JSon(schema_method=`_validator_get`, is_list=False, # default ),
        output = JSon(schema_method=`_validator_return_get`,  is_list=True),
        auth='User'
    )
    def get(self, _id):
        """
        Get partner's informations
        """
        return self._get(_id)

    # Validator
    def _validator_get(self):
        return {}

    def _validator_return_get(self):
        res = self._validator_create()
        res.update({
            'id': {'type': 'integer', 'required': True, 'empty': False},
        })
        return res
```

Serialize / deserialize could be done by the decorator....

lmignon:
fixes #6 and #2 
includes #41


base_rest/models/rest_service_registration.py
```diff
@@ -49,6 +58,52 @@ def _register_hook(self):
             (name, cls) for name, cls in controllers if "RestController" not in name
         ]
         http.controllers_per_module["base_rest"] = controllers
+        # create the final controller providing the http routes for
```
sebastienbeau, line 62: I am testing this new version of rest api, and it's seem that we have some issue here.
If I stop/start odoo and refresh quickly odoo in my web browser it fail to load the registry correctly and my server is crashed for ever, so I need to start and wait some second.
Also If I run the base_rest_demo/test/test_openapi.py I have the issue with the registry. I am not sure that it's related.

@lmignon can you try in local to see if you have the same issue?

sebastienbeau: It's seem that I do not have the issue anymore.

base_rest/restapi.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def get_cerberus_validator(self, service):
+        schema = None
+        if isinstance(self._schema, dict):
+            # schema is a cerberus schema
+            schema = self._schema
+        elif isinstance(self._schema, str):
+            # schema is a method name to call on service to get the schema or
+            schema = getattr(service, self._schema)()
+        if isinstance(schema, Validator):
```
etobella, line 174: I think this needs to be 
```
if isinstance(self._schema, Validator):
    return self._schema
```

etobella (changes requested):
Great Work. Tested and it works perfectly :wink:
I found a small error

base_rest/apispec/base_rest_service_apispec.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def _add_method_path(self, method):
+        description = textwrap.dedent(method.__doc__ or "")
+        routing = method.routing
+        for paths, method in routing["routes"]:
+            for path in paths:
+                self.path(
+                    path,
+                    description=description,
+                    operations={method.lower(): {}},
```
etobella: Change it for:
```
                self.path(
                    path,
                    operations={method.lower(): {'summary': description}},
```
Right now, the description is not shown. With this, it is visible :wink:

base_rest/restapi.py
```diff
@@ -0,0 +1,250 @@
...
+    def __init__(self, name, is_list=False):
+        self._name = name
+        self._is_list = is_list
+
+    def from_params(self, service, params):
+        ModelClass = service.env.datamodels[self._name]
+        try:
+            return ModelClass.load(
+                params, many=self._is_list, unknown=marshmallow.EXCLUDE
+            )
```
etobella: a `.dump()` is required here. It is not working when input json contains a list with a dict inside.

etobella:
@lmignon I am testing your improvement and it is awesome. Much more easier :wink: I am adding the comments with the problems I detected, or do you prefer a PR?

base_rest/restapi.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def from_params(self, service, params):
+        validator = self.get_cerberus_validator(service)
+        if validator.validate(params):
+            return validator.document
+        raise UserError(_("BadRequest %s") % validator.errors)
+
+    def to_response(self, service, result):
+        validator = self.get_cerberus_validator(service)
+        if validator.validate(result):
```
etobella, line 127: Not covered what will happen if it is a list. I have tested with:
```
    def to_response(self, service, result):
        validator = self.get_cerberus_validator(service)
        if self._is_list:
            data = []
            for res in result:
                if not validator.validate(res):
                    raise SystemError(_("Invalid Response %s") % validator.errors)
                data.append(validator.document)
            return data
        if validator.validate(result):
            return validator.document
        raise SystemError(_("Invalid Response %s") % validator.errors)
```
And worked perfectly

base_rest/restapi.py
```diff
@@ -0,0 +1,250 @@
...
+                       cerberus.Validator or a sting with the method name to
+                       call on the service to get the schema or the validator
+        :param is_list:
+        """
+        self._schema = schema
+        self._is_list = is_list
+
+    def from_params(self, service, params):
+        validator = self.get_cerberus_validator(service)
+        if validator.validate(params):
```
etobella, line 121: Not solved what will happen with self._is_list

base_rest/restapi.py
```diff
@@ -0,0 +1,250 @@
...
+    def get_cerberus_validator(self, service):
+        schema = None
+        if isinstance(self._schema, dict):
+            # schema is a cerberus schema
+            schema = self._schema
+        elif isinstance(self._schema, str):
+            # schema is a method name to call on service to get the schema or
+            schema = getattr(service, self._schema)()
+        if isinstance(schema, Validator):
+            return schema
```
lmignon: @etobella IMO the right fix should be:
```suggestion
        schema = self._schema
        if isinstance(self._schema, dict):
            # schema is a cerberus schema
            schema = self._schema
        elif isinstance(self._schema, str):
            # schema is a method name to call on service to get the schema or
            schema = getattr(service, self._schema)()
        if isinstance(schema, Validator):
            return schema
```
We must add a test to cover this case

base_rest/restapi.py
lmignon: `dump()`???? I don't follow you. If your input is a list, you must declare your param's method as `Datamodel('my.rest.model', is_list=1)`

base_rest/apispec/base_rest_service_apispec.py
lmignon: Thank you I've to check

base_rest/restapi.py
lmignon: @etobella Are you sure you use the right version of marshmallow_object and marshamallow? AFAIK is_list is natively supported by the marshmallow_object.... I've to check since I'm pretty sure that should work as is..

lmignon:
@etobella Thank you for your reviews.... I'll try to take a look to your reported issues ASAP but I've a lot of work...

Regards

base_rest/restapi.py
etobella: Good for me

base_rest/restapi.py
etobella: It didn't work on my side. I will check again just to be sure

base_rest/restapi.py
etobella: I know, but it was not received properly on the rest function.

etobella:
If it is not a problem for you, we can make a PR or take the PR from your side (keeping your commits, obviously :smile: )

lmignon:
@etobella I'm not convinced by some of your proposed changes that looks to me like a workaround an other problem (the management of is_list). I would be happy to have this PR merged... but I want to be sure that there is no other to fix the issues you 've encountered. It would be nice to add the required tests to highlight these issues.

Regards,

lmi

etobella:
We can remove the part of is_list if you are not confortable with it

etobella:
According to Cerberus API Documentation (https://docs.python-cerberus.org/en/stable/api.html), the Validator must have an schema and no information about a List is provided on the first element, so you are unable to return `[{'a': 1}, {'a': 2}]`. You can make it with datamodels, not with CerberusValidator

That part was intended to solve this issue, at least on the response part (maybe we shouldn't implement it on the receive, I don't know). However, as I said, if you are not comfortable with it, we can remove it in order to solve it.

base_rest/restapi.py
etobella: My error defining the function. Sorry

lmignon:
@etobella My bad, part of your proposals are for Cerberus not for datamodels. Nevertheless,  the one for the datamodel looks suspicous to me...
@sebastienbeau What's the status of your tests with this new implementation?

etobella:
@lmignon Don't worry, I was capable to solve most of my problems. At the end, the only changes necessary where this ones:

https://github.com/tegin/rest-framework/commit/545c4e1ee05db5212eb723d92751abc86fb72dba

lmignon:
@etobella You need to fix the tests since one of the changes replaces the 'summary' key into openapi document with 'description'
https://travis-ci.com/github/tegin/rest-framework/jobs/441823733#L859

sebastienbeau:
@lmignon I still have this weird issue of missing database in _component_databases[dbname]

This is random I reproduce it locally and also on travis see this job that have failed
https://travis-ci.com/github/shopinvader/odoo-shopinvader/jobs/428039949#L2026
Do you have the issue in local ? Try with the shopinvader PR

Note on shopinvader I have to change some code see my PR here : https://github.com/shopinvader/odoo-shopinvader/pull/802

I make more test soon.

etobella:
@lmignon All tests fixed in my branch, at the end, I made two commits, but they can be squashed :wink: 
https://github.com/tegin/rest-framework/tree/12.0-new-api-wip

lmignon:
@etobella Thank you for all your work. I'll merge your commit. I'll also split base_rest into 3 addons.
* base_rest (core functionalities)
* base_rest_cerberus (base_rest with support for cerberus message definitions)
* base_rest_datamodels (base_rest with support for datamodels message bindings)

@simahawk, @sebastienbeau  What do you thing about the split proposal?

PR #49, opened 2019-10-17
[10.0] black + pre-commit

lmignon:


lmignon:
Fast tracking: only cosmetic changes

PR #50, opened 2019-10-17
[11.0] black + pre-commit

lmignon:
forward port of #49

PR #51, opened 2019-10-17
[12.0]  black + pre-commit

lmignon:
forward port of #50 + black isort on graphql addons


lmignon:
Fast tracking: only cosmetic changes

PR #58, opened 2020-02-13
Allow disabling a service when _is_rest_service_component is False

guewen:
Currently, only the presence of the attribute is inspected, which means

  _is_rest_service_component = False

Has the same effect as:

  _is_rest_service_component = True

With this change, the former will not serve the component in the REST API.

### DIFF
--- a/base_rest/controllers/api_docs.py
+++ b/base_rest/controllers/api_docs.py
@@ -62,7 +62,7 @@ class ApiDocsController(Controller):
     def _filter_service_components(self, components):
         r = []
         for c in components:
-            if hasattr(c, "_is_rest_service_component") and c._usage:
+            if getattr(c, "_is_rest_service_component", None) and c._usage:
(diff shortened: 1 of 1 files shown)


lmignon (approved):
Makes sens. Thank you @guewen

lmignon:
@guewen you should take a look at the new api... https://github.com/OCA/rest-framework/pull/48/files#diff-446e66ad580b33ff05f11e44b939cd38R7 The implementation of REST services becomes smoother, cleaner, and a lot more flexible... With this new API you can define your own url path with query parameters, date in/out are python object, odoo controllers are generated ...

guewen:
Thanks! I'll have a look

lmignon:
little fix -> fast tracking...
/ocabot merge patch

marcelsavegnago:
v12.0 https://github.com/OCA/rest-framework/pull/128

PR #61, opened 2020-03-17
[FIX] Usage mention correct /api-docs discovery URL

dreispt:


lmignon (approved):
Thank you @dreispt !

dreispt:
I think this is good to merge.

PR #62, opened 2020-03-25
[IMP] Log errors on missing validators

dreispt:
This will help putting struggling developers in the right path.

### DIFF
--- a/base_rest/components/service.py
+++ b/base_rest/components/service.py
@@ -93,6 +93,7 @@ class BaseRestService(AbstractComponent):
 
     def _get_validator(self, validator_method):
         if not hasattr(self, validator_method):
+            _logger.error("Missing validator %s", validator_method)
             return None
         v = getattr(self, validator_method)()
         if isinstance(v, dict):


base_rest/components/service.py
```diff
@@ -93,6 +93,7 @@ def _log_call(self, func, params, secure_params, res):
 
     def _get_validator(self, validator_method):
         if not hasattr(self, validator_method):
+            _logger.error("Missing validator %s", validator_method)
```
lmignon, line 96: @dreispt Validators are only required for input messages. An error is raised or log.warning is called into the _secure_input and _secure_output methods. IMO this change is useless.

dreispt: This might not be the best fix, but a missing output validator does not log a warning, it raises a traceback that does not give specific information on what expected validator is missing, sending the develop down into a debug rabbit hole.

sbidoul: > a missing output validator does not log a warning, it raises a traceback

@dreispt That is surprising. Can you provide more details? There is code to log a warning in that case:

https://github.com/OCA/rest-framework/blob/28a8436dea7fbf69b5796948ccdc5d2df56e5c0c/base_rest/components/service.py#L161-L169

dreispt: IIRC that is true only for missing input validators, not for output validator. But I'll double check.

PR #64, opened 2020-05-08
base_rest: call openapi methods only once

simahawk:


simahawk:
ouch! :cry:  I pushed to 13.0 straight w/ an amended commit as well due to mis-configuration on my remotes.
I can revert if not accepted anyway.

simahawk:
I've reverted the commit on 13.0 as it fails. I check later here

simahawk:
https://github.com/OCA/rest-framework/pull/65

PR #65, opened 2020-05-08
 base_rest: call openapi methods only once

simahawk:
Pointless to call them on every loop
since the result won't change.

Replaces  #64

### DIFF
--- a/base_rest/components/service.py
+++ b/base_rest/components/service.py
@@ -290,6 +290,9 @@ class BaseRestService(AbstractComponent):
                 continue
             public_methods[name] = data
 
+        default_parameters = self._get_openapi_default_parameters()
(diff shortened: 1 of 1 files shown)


simahawk:
https://github.com/OCA/rest-framework/pull/65

lmignon (approved):

lmignon:
> #65

Thank you @simahawk 
I hope to replace the current implementation with a more flexible new approach (keeping the backward compatibility with the current one)

If interested, You can take a look at https://github.com/OCA/rest-framework/pull/48
https://github.com/OCA/rest-framework/pull/48/files#diff-446e66ad580b33ff05f11e44b939cd38R7

Regards,

lmi

sbidoul:
Should be green now, following #75. Let's see

/ocabot merge patch

PR #66, opened 2020-05-12
[12.0] graphql_base: Add aliases and handle 2one relations better

janverb:
I found a few ways to reduce boilerplate when working with `graphql_base`.

When only one record is expected, empty recordsets can be replaced by `None`, so you don't need a custom resolver to handle that.

When you just want to rename a field (e.g. `partner_id` to `partner`) the new `alias()` function can create an appropriate resolver for you. That feels like it could be a feature in `graphene` but I couldn't find anything.

### DIFF
--- a/graphql_base/types.py
+++ b/graphql_base/types.py
@@ -72,8 +72,7 @@ def alias(target_name):
     @classmethod
     def resolve_field(cls, root, info):
         resolver = (
-            cls._meta.default_resolver
-            or graphene.types.resolver.get_default_resolver()
+            cls._meta.default_resolver or graphene.types.resolver.get_default_resolver()
         )
         field = cls._meta.fields.get(info.field_name, None)
         default_value = getattr(field, "default_value", None)
--- a/graphql_base/types.py
+++ b/graphql_base/types.py
@@ -3,7 +3,7 @@
 
 import graphene
 
-from odoo import fields
+from odoo import fields, models
 
 
 def odoo_attr_resolver(attname, default_value, root, info, **args):
@@ -16,6 +16,8 @@ def odoo_attr_resolver(attname, default_value, root, info, **args):
 
     It converts datetimes to the user timezone.
 
+    It converts a lack of a record to None, if a single record is expected.
+
     It also raises an error if the attribute is not present, ignoring
     any default value, so as to return if the schema declares a field
     that is not present in the underlying Odoo model.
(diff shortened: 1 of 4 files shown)


graphql_base/types.py
```diff
@@ -41,3 +49,33 @@ def __init_subclass_with_meta__(cls, default_resolver=None, **options):
...
+    """
+
+    @classmethod
+    def resolve_field(cls, root, info):
+        resolver = (
+            cls._meta.default_resolver or graphene.types.resolver.get_default_resolver()
+        )
+        field = cls._meta.fields.get(info.field_name, None)
+        default_value = getattr(field, "default_value", None)
+        return resolver(target_name, default_value, root, info)
```
sbidoul, line 79: Could this be simplified to something like this?

```suggestion
    @staticmethod
    def resolve_field(root, info):
        return getattr(root, target_name) or None
```

graphql_base/types.py
```diff
@@ -27,6 +29,12 @@ def odoo_attr_resolver(attname, default_value, root, info, **args):
             return None
     elif isinstance(field, fields.Datetime):
         return fields.Datetime.context_timestamp(root, value)
+    elif (
+        not isinstance(field, fields._RelationalMulti)
+        and isinstance(value, models.BaseModel)
+        and not value
+    ):
```
sbidoul, line 36: I do not immediately see in which case this part applies. Could you explain a little bit?

sbidoul (commented):
Hi, thanks for this very interesting contribution. I've a couple of questions.

graphql_base/types.py
janverb: It copies the way graphene looks for values, so you still get all the finicky things `odoo_attr_resolver` does by default. And if you use your own object type with other custom behavior then that's handled too.

`getattr(root, target_name) or None` would work for the example in the docstring, but not for datetimes, 2many relations, and so on.

graphql_base/types.py
janverb: It's to detect that it's a 2one relationship without a value. Those should return `None`.

If it's a 2many relationship (`isinstance(field, fields._RelationalMulti)`) then returning an empty recordset is fine, that'll get turned into an empty array like it should. (It might have been better to instead check in the schema if the field type is an array, but I couldn't figure out how to do that properly.)

And if it's not a recordset (`isinstance(value, models.BaseModel)`) then it's not a relation so the whole thing doesn't apply.

janverb:
I no longer work with this but @bamilab could possibly take it over.

PR #67, opened 2020-05-25
fix typo in attribute

gurneyalex:


lmignon (approved):
Thank you @gurneyalex

PR #70, opened 2020-05-25
fix typo in attribute

gurneyalex:


lmignon (commented):
Thank you @gurneyalex

lmignon (approved):

PR #76, opened 2020-08-26
base_rest: improve exception wrapped info

simahawk:
Your rest app can now raise an exception that provide 'rest_json_info' attribute.
This attribute must be a dict and should contain relevant - but not sensitive - info
on the error occured (eg: a link to a tracking system w/ a reference to the error).

PR #80, opened 2020-08-27
base_rest: pick exception info from all exc types

simahawk:
Evolution of #76. You could have custom exceptions inheriting from those specific exceptions. This makes them work all in the same way.

### DIFF
--- a/base_rest/http.py
+++ b/base_rest/http.py
@@ -172,13 +172,19 @@ class HttpRestRequest(HttpRequest):
         try:
             return super(HttpRestRequest, self)._handle_exception(exception)
         except (UserError, ValidationError) as e:
-            return wrapJsonException(BadRequest(e.name), include_description=True)
+            extra_info = getattr(e, "rest_json_info", None)
+            return wrapJsonException(
+                BadRequest(e.name), include_description=True, extra_info=extra_info
+            )
         except MissingError as e:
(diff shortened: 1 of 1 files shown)


base_rest/http.py
```diff
@@ -172,13 +172,19 @@ def _handle_exception(self, exception):
         try:
             return super(HttpRestRequest, self)._handle_exception(exception)
         except (UserError, ValidationError) as e:
-            return wrapJsonException(BadRequest(e.name), include_description=True)
+            extra_info = getattr(e, "rest_json_info", None)
```
lmignon, line 175: @simahawk To avoid code redundancy we could put at the start of the `wrapJsonException`method:
```python
def wrapJsonException(exception, include_description=False, extra_info=None):
    """Wrap exceptions to be rendered as JSON.
    :param exception: an instance of an exception
    :param include_description: include full description in payload
    :param extra_info: dict to provide extra keys to include in payload
    """
    extra_info = extra_info or getattr(exception, "rest_json_info", None)
    
``` 
What do you think about?

simahawk: you should pass the original exception there (I thought about it, decided to keep it simple) because `exception` in this case is the wrapping one.

lmignon: indeed....

lmignon (approved):

PR #71, opened 2020-05-26
[13.0] base_rest: New api

lmignon:
Foward port of #48

PR #88, opened 2020-11-23
[14.0][MIG][WIP] base_rest, base_rest_demo: Migration to 14.0

kevinkhao:
Depends on https://github.com/OCA/rest-framework/pull/71


lmignon (approved):
Thank you for the migration @kevinkhao

lmignon:
@kevinkhao I propose to initialize the 14.0 branch from the 13.0 and declare all the add-on uninstal-able . @sbidoul Do you agree with this process.

sbidoul:
> I propose to initialize the 14.0 branch from the 13.0 and declare all the add-on uninstal-able . @sbidoul Do you agree with this process.

Fine with me.

kevinkhao:
@lmignon sure. For what I did, it's still WIP as I will test it soon in a real scenario

kevinkhao:
@lmignon maybe I could do it; but because I've never done this before, I need verbose details (including the obvious ones) on what is the best way. For example, would you just throw away the current OCA 14.0 ? You want to use 13.0-48-port, apply Copier procedure, and declare that as the new 14.0 OCA branch ?  
Or you mean throwing away OCA 14.0, and creating a new OCA 14.0 from acsone/13.0 ?
I have tried going from 14.0 and doing the format-patch using oca/14.0..acsone/13.0-48-port but there are a lot of conflicts which I am not sure about.

PR #97, opened 2021-01-19
[14.0][FIX] base_rest: Allows to scroll in case of overflow into the swagge…

### LINKED ISSUE #94: Swagger UI accessible through REST API menuitem is broken
Swagger UI seems broken as it is impossible to scroll.

image

lmignon:
…r ui

Before this change, since the assets_frontend was loaded into the swagger ui the overflow was hidden. It was therfore no more possible to scroll into the swagger UI in case of overflow.
The dependency can be safely removed since it's not used by the swagger UI

fixes #94


lmignon:
ping @kevinkhao

PR #99, opened 2021-01-20
[13.0] Add handlers to delegate validators

guewen:
CerberusValidator looks for the method name on the service component to
get the schema validators.
Delegate this to a Component, so we can customize the way the validator
for a service is found.

E.g.:

* for each service component, we have 2 additional components: a
  validator for the input schema and another for the output schema
* when looking for the validator, we have to search for the
  corresponding input/output validation component

As implemented in: OCA/wms#137

By delegating this lookup to a base "cerberus.validator" component, we
can override the default behavior for a services collection, by
inheriting the Component.

By default, the behavior does not change.

### DIFF
--- a/base_rest/models/rest_service_registration.py
+++ b/base_rest/models/rest_service_registration.py
@@ -237,17 +237,21 @@ class RestApiMethodTransformer(object):
 
         return routes
 
-    def _method_to_input_param(self, method):
-        validator_method_name = "_validator_{}".format(method.__name__)
-        if hasattr(self._service, validator_method_name):
+    def _method_to_param(self, validator_method_name):
+        validator_component = self._service.component(usage="cerberus.validator")
+        if validator_component.has_validator_handler(
+            self._service, validator_method_name
+        ):
             return restapi.CerberusValidator(schema=validator_method_name)
         return None
 
+    def _method_to_input_param(self, method):
+        validator_method_name = "_validator_{}".format(method.__name__)
+        return self._method_to_param(validator_method_name)
+
     def _method_to_output_param(self, method):
         validator_method_name = "_validator_return_{}".format(method.__name__)
-        if hasattr(self._service, validator_method_name):
-            return restapi.CerberusValidator(schema=validator_method_name)
-        return None
+        return self._method_to_param(validator_method_name)
 
 
 class RestApiServiceControllerGenerator(object):
--- a/base_rest/models/rest_service_registration.py
+++ b/base_rest/models/rest_service_registration.py
@@ -237,21 +237,21 @@ class RestApiMethodTransformer(object):
 
         return routes
 
-    def _method_to_param(self, validator_method_name):
+    def _method_to_param(self, validator_method_name, direction):
         validator_component = self._service.component(usage="cerberus.validator")
         if validator_component.has_validator_handler(
-            self._service, validator_method_name
+            self._service, validator_method_name, direction
         ):
             return restapi.CerberusValidator(schema=validator_method_name)
         return None
 
     def _method_to_input_param(self, method):
         validator_method_name = "_validator_{}".format(method.__name__)
-        return self._method_to_param(validator_method_name)
+        return self._method_to_param(validator_method_name, "input")
 
     def _method_to_output_param(self, method):
         validator_method_name = "_validator_return_{}".format(method.__name__)
-        return self._method_to_param(validator_method_name)
+        return self._method_to_param(validator_method_name, "output")
 
 
 class RestApiServiceControllerGenerator(object):
--- a/base_rest/components/__init__.py
+++ b/base_rest/components/__init__.py
@@ -1,1 +1,2 @@
 from . import service
+from . import cerberus_validator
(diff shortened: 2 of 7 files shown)


guewen:
I'll add an example very soon.

guewen:
Some tests fail because of this:

```2021-01-20 11:18:13,486 5786 ERROR openerp_test odoo.addons.base_rest.tests.test_cerberus_validator: ERROR: TestCerberusValidator.test_schema_lookup_from_string

File "/home/travis/build/OCA/rest-framework/setup/base_rest/odoo/addons/base_rest/tests/test_cerberus_validator.py", line 276, in test_schema_lookup_from_string
validator = v.get_cerberus_validator(MyService())
File "/home/travis/build/OCA/rest-framework/setup/base_rest/odoo/addons/base_rest/restapi.py", line 173, in get_cerberus_validator
schema = service._get_validator_handler(self._schema)()
AttributeError: 'MyService' object has no attribute '_get_validator_handler'
```

The thing is:
* in the test, an almost bare object (`MyService`) is sent to `CerberusValidator.get_cerberus_validator()`, so it doesn't have the `_get_validator_handler` and `_has_validator_handler` methods
* in the implementation of `base_rest`, `CerberusValidator.get_cerberus_validator()` is always called with a service `base.rest.service` or inheriting from it, the `_get_validator_handler` and `_has_validator_handler` methods are implemented there

What should be accepted interface for `CerberusValidator.get_cerberus_validator()`? Should I add the 2 handler methods in the tests class `MyService` to conform to the implemented service interface, or should I check for the presence of the handler methods in `CerberusValidator.get_cerberus_validator()`?

base_rest/components/service.py
```diff
@@ -175,6 +175,23 @@ def _validator_get(self):
         """
         return {}
 
+    def _get_validator_handler(self, method_name):
```
lmignon: @guewen I would like to avoid to re introduce some logic dedicated to the message processing into  `base.rest.service`. 
Instead of adding these new methods, could we not consider to look for a component in the context of the service with usage='cerberus.validator' ? The `RestApiMethodTransmer` and `CerberusValidator`classes could lookup the registry to get this component and you could extend it into shopfloor :thinking:

guewen: This is a better separation of concern, yes.

guewen:
@lmignon would my new implementation be more aligned with your vision?
I still have an issue with `MyService` in tests as I can't use `self._service.component` on it in `RestApiMethodTransformer`.

base_rest/models/rest_service_registration.py
```diff
@@ -237,17 +237,21 @@ def _method_to_routes(self, method):
 
         return routes
 
-    def _method_to_input_param(self, method):
-        validator_method_name = "_validator_{}".format(method.__name__)
-        if hasattr(self._service, validator_method_name):
+    def _method_to_param(self, validator_method_name):
+        validator_component = self._service.component(usage="cerberus.validator")
```
simahawk, line 241: can't we look for a specific usage by direction instead? eg: `cerberus.validator.[input|output]`?
Then in [_get_validator_component](https://github.com/OCA/wms/pull/137/files#diff-4eac90c52a977b0f3aa9e5564668b83ee346d4d1d4f8e57452346d4986d6cfc4R18) you don't have to check.

simahawk: ah but then we'll end up w/ too many components... never mind

guewen: You don't have the direction in `CerberusValidator` (https://github.com/OCA/rest-framework/pull/99/files#diff-232b42925d810d12d892cb1d674234d420568553ccd871838eb30f9541eed8ccR172), maybe this is something we can add, what's your opinion @lmignon?

simahawk: even if not used in this case I'd say that the CerberusValidator instance should know the direction (eg: `self._direction = "input|output"`)

lmignon: `RestMethodParam` is used for the input|ouput direction by providing methods to use according to the direction `from_params` and `to_response`.
I hesitated for a long time to create a class by direction and I finally took the approach of grouping in the same class. Perhaps this is not the best approach.
It could make sense to have a component by direction but aren't we going to complicate things too much?

lmignon:
> RestApiMethodTransformer

base_rest/models/rest_service_registration.py
guewen: A single component seems better as if we have 2 components, they'd probably have to share some logic, and it makes the big picture too complex.

Can we add a "direction" parameter to `get_cerberus_validator` and `to_json_schema` that would be passed from `from_params` / `to_response` / `to_openapi_responses` / `to_openapi_requestbody` to the new component?

lmignon: > Can we add a "direction" parameter to `get_cerberus_validator` and `to_json_schema` that would be passed from `from_params` / `to_response` / `to_openapi_responses` / `to_openapi_requestbody` to the new component?

For sure @guewen

guewen:
Added a direction parameter

lmignon (approved):
Thank you @guewen LGTM (Code review + Functional tests)

lmignon:
ping @simahawk

simahawk (approved):

guewen:
Thanks for your reviews and ideas!

lmignon:
you're welcome :smirk:

PR #100, opened 2021-01-20
python accept_language module not found

andreampiovesana:
is parse-accept-language?


lmignon:
I'm a little bit surprised since we have no error on travis nor runbot...
Nevertheless your change makes sense but if we put the package name into the manifest, we should also modify the external-dependencies-override into setup/base_rest/setup.py

PR #101, opened 2021-01-22
[12.0] Add handlers to delegate validators

lmignon:
backport of #99


lmignon:
/ocabot merge patch

Fast tracking since #99 is approved and merged

PR #106, opened 2021-02-01
[13.0] Add rest_log

simahawk:
moved out from Shopfloor app.


simahawk:
@lmignon @guewen tests fail because the logging is always [enabled by default](https://github.com/OCA/rest-framework/pull/106/files#diff-cc8dc8b0b0f1883d628bc7d3acb73d5dd3537ceaad2c1b14cf6997cfccbeb8d3R39). IMO we should have an explicit enable flag which you can turn on via setting. Maybe:

- add an ir.config.param to enable it by collection (the content can be a CSV of collection names)
- `_log_calls_in_db` is false by default but can be used to force enabling

what do you think?


BTW I'm thinking since a while that we should have a rest backend where to configure thing per service (eg: allowed api keys for a specific service, enable logging, enable special params for specific collections, etc).

simahawk:
@lmignon @guewen gentle ping :innocent:

rest_log/components/service.py
```diff
@@ -0,0 +1,148 @@
...
+
+from werkzeug.urls import url_encode, url_join
+
+from odoo import exceptions, registry
+from odoo.http import request
+
+from odoo.addons.component.core import AbstractComponent
+
+
+class RESTServiceDispatchException(Exception):
```
lmignon: @simahawk I would prefer to have the exeception classes defined into `exceptions.py`

rest_log/components/service.py
```diff
@@ -0,0 +1,148 @@
...
+        return self._dispatch_with_db_logging(method_name, *args, params=params)
+
+    def _db_logging_active(self):
+        return (
+            request and self._log_calls_in_db and self.env["rest.log"].logging_active()
+        )
+
+    def _dispatch_with_db_logging(self, method_name, *args, params=None):
+        try:
+            result = super().dispatch(method_name, *args, params=params)
```
lmignon, line 37: @simahawk would be it possible to use a savepoint here? That would avoid the call to rollback and the creation of a new cursor to log the error. :thinking: 

```python

def _dispatch_with_db_logging(self, method_name, *args, params=None):
    with self._process_exception(self.env, method_name, *args, params),  self.env.cr.savepoint()
        return super().dispatch(method_name, *args, params=params)

@contextmanager
def pro._process_exception(self, env, method_name, *args, params=None):
    try:
        yield
    except Exception e:
        # At this stage the roolback is already done by the savepoint...
        # here we log the exception
        ....
        # and we build a werkzeug response for the exception using the same logic as the one in odoo.addons.base_rest.http -> to be refactored?
        from odoo.http import request
        except_wrapper = request._handle_exception(e)
        headers = except_wrapper.get_body(env)
        body = except_wrapper.get_headers(env)
        return request.make_json_response(body, headers)
```
```

File: rest_log/components/service.py [hunk 3]
```diff
@@ -0,0 +1,148 @@
...
+            )
+        except Exception as orig_exception:
+            self._dispatch_exception(
+                RESTServiceDispatchException, orig_exception, *args, params=params
+            )
+        log_entry = self._log_call_in_db(
+            self.env, request, *args, params=params, result=result
+        )
+        log_entry_url = self._get_log_entry_url(log_entry)
+        result["log_entry_url"] = log_entry_url
```
lmignon: @simahawk A service can also return a werkzeug response (to return a binary content for exemple). This line will not work if the result of the call to the service method is not a dict

rest_log/components/service.py
```diff
@@ -0,0 +1,148 @@
...
+    def _db_logging_active(self):
+        return (
+            request and self._log_calls_in_db and self.env["rest.log"].logging_active()
+        )
+
+    def _dispatch_with_db_logging(self, method_name, *args, params=None):
+        try:
+            result = super().dispatch(method_name, *args, params=params)
+        except exceptions.UserError as orig_exception:
+            self._dispatch_exception(
```
lmignon, line 39: @simahawk What's the plus value of wrapping the exception into another exception ???

rest_log/hooks.py
```diff
@@ -0,0 +1,21 @@
...
+
+def pre_init_hook(cr):
+    """On first install copy recods from shopfloor_log table if available.
+
+    """
+    cr.execute("SELECT 1 FROM pg_class WHERE relname = 'rest_log'")
+    if cr.fetchone():
+        # rest_log was already installed
+        return
+    cr.execute("SELECT 1 FROM pg_class WHERE relname = 'shopfloor_log'")
```
lmignon, line 16: ?shopfloor?

lmignon (commented):
Thank you @simahawk I wonder about the performance cost of logging all the REST calls. Comments into the code...

sbidoul:
@simahawk regarding configurability, I'd start with an ir.config.param to enable logging explicitly. Since the only current users of this module are happy to log all calls, I'd say more advanced configurability (by collection or with a full fledged configuration object) could be left for future work if/when the need arises.

rest_log/components/service.py
simahawk: seems a good idea, I'll check, tnx!

rest_log/components/service.py
simahawk: IMO in that case you should have binary data wrapped in a dict which may contain metadata.
But yeah, I'll add a check.

rest_log/components/service.py
simahawk: The idea was to have always a "good" exception class which can carry all the info we need and that could simply `wrapJsonException` in the future.

rest_log/hooks.py
simahawk: well, this module comes from there and I wanted to preserve old logs for shopfloor users. This is the only safe/sane way I've found w/out making it too complicated. Being just a query there's no tight dependency and I think is a good compromise.

simahawk:
> @simahawk regarding configurability, I'd start with an ir.config.param to enable logging explicitly. Since the only current users of this module are happy to log all calls,

yes, I agree.

> I'd say more advanced configurability (by collection or with a full fledged configuration object) could be left for future work if/when the need arises.

Of course! This was more to start thinking/discussing about this option for the future. The work here simply reminded me that I want to talk about this since a while :wink: 
I'll track it in an issue!

rest_log/hooks.py
lmignon: @simahawk Not a problem for me :smirk:

rest_log/components/service.py
simahawk: @lmignon I added a TODO for this. I'd prefer to keep as it is right now. Fine for you?

simahawk:
enable by config param added!

> Thank you @simahawk I wonder about the performance cost of logging all the REST calls. Comments into the code...

We haven't faced any perf issue till today. Actually for the value it gives you when testing a new system I think is a good trade off.

rest_log/components/service.py
lmignon: OK, even if I'm not a fan of code where we create a new cursor and make explicit call to `rollback` :smirk:

simahawk: @guewen do you remember any special reason to do it this way?

guewen: No, I suppose it can be replaced by a savepoint. If we can guarantee that nothing has been written to db before the call to `super().dispatch(method_name, *args, params=params)`.

rest_log/models/rest_log.py
```diff
@@ -0,0 +1,130 @@
...
+        # something broken somewhere
+        "ValueError": "severe",
+        "AttributeError": "severe",
+        "UnboundLocalError": "severe",
+    }
+
+    request_url = fields.Char(readonly=True, string="Request URL")
+    request_method = fields.Char(readonly=True)
+    params = fields.Text(readonly=True)
+    # TODO: make these fields serialized and use a computed field for displaying
```
guewen, line 31: wouldn't you lose search capabilities?

guewen:
As it adds a lot of methods regarding logs in the base service, maybe we should delegate this to a "log" component as I did for the cerberus validator in https://github.com/OCA/rest-framework/pull/99/files#diff-232b42925d810d12d892cb1d674234d420568553ccd871838eb30f9541eed8ccR175? (or move the methods to the model itself).

lmignon:
> As it adds a lot of methods regarding logs in the base service, maybe we should delegate this to a "log" component as I did for the cerberus validator in https://github.com/OCA/rest-framework/pull/99/files#diff-232b42925d810d12d892cb1d674234d420568553ccd871838eb30f9541eed8ccR175? (or move the methods to the model itself).

I personally would avoid using a component for log processing. I have the feeling that this would add an indirection in the reading of the code that is not necessary.

rest_log/components/service.py
simahawk: > No, I suppose it can be replaced by a savepoint. If we can guarantee that nothing has been written to db before the call to `super().dispatch(method_name, *args, params=params)`.

Can we go ahead as it is now?
I'd prefer to not touch a part that we know is working fine for thousands of requests today.
We have all the time to improve this bit w/out affecting dependencies.
ATM this is blocking (or making more complicated) split PRs https://github.com/OCA/wms/pull/156 and https://github.com/OCA/wms/pull/160 (as well as v14 work).

simahawk:
> As it adds a lot of methods regarding logs in the base service, maybe we should delegate this to a "log" component as I did for the cerberus validator in https://github.com/OCA/rest-framework/pull/99/files#diff-232b42925d810d12d892cb1d674234d420568553ccd871838eb30f9541eed8ccR175? (or move the methods to the model itself).

A component seems too much, the model seems not the right place.
As per https://github.com/OCA/rest-framework/pull/106#discussion_r573743673 I'd go as it is for now.

guewen:
> A component seems too much, the model seems not the right place.
> As per [#106 (comment)](https://github.com/OCA/rest-framework/pull/106#discussion_r573743673) I'd go as it is for now.

My comment was according to the comment in https://github.com/OCA/rest-framework/pull/99#discussion_r560927203, but I missed the "dedicated to the message processing".
Fine with me either way.

guewen (approved):

lmignon (approved):
Let's move forward... Thank you @simahawk

PR #109, opened 2021-02-09
[11.0][FIX] Check database existence before initializing it

### LINKED ISSUE #108: Odoo crashes after dropping a database
It might happen that a database is deleted, expecially while developing new features.
If module `base_rest` is installed in at least one DB (even the deleted one), the `get_request` override in `base_rest/http.py` will eventually try to initialize the deleted database without checking if it still exists. So when the `odoo.registry(db)` line is reached, Odoo tries to initialize a Registry for a non existing database, and it crashes.

PRs:

- [ ] 11.0: #109
- [x] 12.0: #110
- [x] 13.0: #111
- [x] 14.0: #112

SilvioGregorini:
Closes #108

### DIFF
--- a/base_rest/http.py
+++ b/base_rest/http.py
@@ -191,7 +191,7 @@ ori_get_request = Root.get_request
 
 def get_request(self, httprequest):
     db = httprequest.session.db
-    if db:
+    if db and odoo.service.db.exp_db_exist(db):
         # on the very first request processed by a worker,
         # registry is not loaded yet
         # so we enforce its loading here to make sure that


matteoopenf (approved):

lmignon:
Thank you @SilvioGregorini
We have to fix travis in order to merge your PR.

SilvioGregorini:
@lmignon in the meanwhile, should I open the same PR for other versions of the module (12, 13, 14)?

lmignon:
> @lmignon in the meanwhile, should I open the same PR for other versions of the module (12, 13, 14)?

For sure

matteoopenf:
Is it possible merge?

simahawk:
already merged on next versions but still red here.
@SilvioC2C can you have a look?

SilvioC2C:
@simahawk apparently we're having a dependency conflict issue with `pre-commit`:
```
    The conflict is caused by:
        pylint-odoo 3.5.0 depends on pylint==2.5.3; python_version >= "3"
        pylint-plugin-utils 0.6 depends on pylint>=1.7
        pre-commit-dummy-package 0.0.0 depends on pylint==2.3.1
    
    To fix this you could try to:
    1. loosen the range of package versions you've specified
    2. remove package versions to allow pip attempt to solve the dependency conflict
```

PR #111, opened 2021-02-10
[13.0][FIX] Check database existence before initializing it

### LINKED ISSUE #108: Odoo crashes after dropping a database

SilvioGregorini:
Closes #108

### DIFF
--- a/base_rest/http.py
+++ b/base_rest/http.py
@@ -211,7 +211,7 @@ ori_get_request = Root.get_request
 
 def get_request(self, httprequest):
     db = httprequest.session.db
-    if db:
+    if db and odoo.service.db.exp_db_exist(db):
         # on the very first request processed by a worker,
         # registry is not loaded yet
         # so we enforce its loading here to make sure that


lmignon (approved):

base_rest/http.py
```diff
@@ -211,7 +211,7 @@ def make_json_response(self, data, headers=None, cookies=None):
 
 def get_request(self, httprequest):
     db = httprequest.session.db
-    if db:
+    if db and odoo.service.db.exp_db_exist(db):
```
sbidoul, line 214: Calling `exp_db_exist` at the beginning of each request could have significant performance impacts ?

SilvioGregorini: @sbidoul actually, I didn't run any test about it. Got any idea about how to check it?

len-foss: You could run a script to make a significant number of requests with curl and profile performance (https://www.odoo.com/documentation/14.0/howtos/profilecode.html).

If the impact is not negligible this could simply be put into another module that just overrides get_request to make the check, and it might depend on server-env to be skipped in prod <insert joke about how often prod db are dropped>.

SilvioGregorini: @sbidoul @len-foss I tested both version of the method, confronting the execution time for both cases (whether `exp_db_exist` is called or not), testing batches of 1, 10, 100, 1000, 10000 requests.

The difference is totally negligible for batches up to 1000 requests (< 0.5% delay); with 1000 requests, I found an average of 1.5% delay (which means, for example, that 1000 requests are managed in 82.809 seconds when `exp_db_exist` is not called, 84.041 seconds otherwise), and it grows up to ~2.7% for 10k requests (1796.228s without `exp_db_exist`, 1843.874s with it).

What do you think of it?

PR #115, opened 2021-02-16
rest_log: fix init hook

simahawk:
Previous hook was cloning the table
but by doing so SQL sequences and indexes were missing.


simahawk:
@lmignon if you don't have anything against it I'd fast track it as it fixes the transition from shopfloor_log.

lmignon (approved):

PR #117, opened 2021-02-18
[BUG] issue with datamodel, recursion limit

sebastienbeau:
@lmignon @sbidoul 
It seem that we have a huge issue with datamodel.

If you request a lot of record (around 1000) you will have a limit recursion error.
Also (real case in pre-production) when you call again and again the same api you also have the erreur (the limit is cumulative as the object still exist in the registry).

The error is in Mashmallow lib ...

### DIFF
--- a/base_rest_demo/tests/__init__.py
+++ b/base_rest_demo/tests/__init__.py
@@ -1,3 +1,4 @@
 from . import test_controller
 from . import test_openapi
 from . import test_exception
+from . import test_new_api
--- a/base_rest_demo/tests/common.py
+++ b/base_rest_demo/tests/common.py
@@ -5,8 +5,10 @@ import json
 import os
 
 from odoo.addons.base_rest.controllers.main import _PseudoCollection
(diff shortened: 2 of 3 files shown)


etobella:
:astonished: I can confirm the issue, however, it is hard to know how to solve it :thinking:

lmignon:
Unfortunately I've no time to dig into this issue ATM

etobella:
Adding this test
```
    def test_search2(self):
        self.env["res.partner"].create({"name": "P"})
        for _idx in range(0, 1000):
            print(_idx)
            self.service.dispatch("search")
```
At the 24 the error is raised :exploding_head:

etobella:
Hi, I think I found a way to solve the issue, I am still working on it as the tests of `datamodel` are failing due to the change :smile: 

Do you want a new PR or a PR directly on your repo?

etobella:
@sebastienbeau

lmignon:
@sebastienbeau I close this one since it's solved by #120 I'll do the forward port in 13 and 14

Issue #116, opened 2021-02-17
maximum recursion error with new api 


scrimo:
Hy! 

I've made a REST endpoint similar to the base_rest_demo. i search products and create from odoo product.product object an datamodel object.

this is my datamodel:
`
from marshmallow import fields

from odoo.addons.datamodel.core import Datamodel
#from odoo.addons.datamodel.fields import NestedModel

class SCDatamodelProduct(Datamodel):
    _name = "sc.datamodel.product"

    id = fields.Integer(required=True, allow_none=False)
    name = fields.String(required=True, allow_none=False)
    price = fields.Float(required=True, allow_none=False)
    taxes_id = fields.Integer(allow_none=True)
    tax_included = fields.Boolean(required=True, allow_none=False)
    tax_percentage = fields.Float(allow_none=False, default=0.0)
    barcode = fields.String(allow_none=True)
    internal_reference = fields.String(allow_none=True)
    qty_on_hand = fields.Float(allow_none=False, default=0.0)
    category = fields.String(allow_none=True)
    product_groups = fields.String(allow_none=True)
    size_list = fields.String(allow_none=True)
    style_list = fields.String(allow_none=True)
    weight = fields.String(allow_none=True)`

first time when calling the enpoint, everything works correctly. but at all following calls i get an error:

[...]

PR #118, opened 2021-02-19
[FIX] issue with datamodel, recursion limit: 12.0

etobella:
This PR intends to fix the issue found on #116  and #117 
@lmignon @sebastienbeau @sbidoul

### DIFF
--- a/datamodel/core.py
+++ b/datamodel/core.py
@@ -2,7 +2,6 @@
 # Copyright 2019 ACSONE SA/NV
 # License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
 
-import functools
 import logging
 from collections import OrderedDict, defaultdict
 
@@ -93,7 +92,7 @@ _datamodel_databases = DatamodelDatabases()
 
 @marshmallow.post_load
 def __make_object__(self, data, **kwargs):
-    datamodel = self._env.datamodels[self._datamodel_name]
+    datamodel = self._registry[self._datamodel_name]
     return datamodel(__post_load__=True, __schema__=self, **data)
 
 
@@ -192,11 +191,6 @@ class Datamodel(MarshmallowModel, metaclass=MetaDatamodel):
     def __init__(self, context=None, partial=None, env=None, **kwargs):
         self._env = env
         super().__init__(context=context, partial=partial, **kwargs)
-
-    @property
-    def env(self):
-        """ Current datamodels registry"""
-        return self._env
 
     @classmethod
     def get_schema(cls, **kwargs):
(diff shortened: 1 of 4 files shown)


etobella:
The way to solve the issue was removing the dependency on Environment inside the Datamodel, using the registry was enough and the registry was already inside.

etobella:
@sebastienbeau Did you test the PR in your environment? Did it work properly?

lmignon:
@etobella By removing the access to Environment from inside the Datamodel you remove the ability to access to the environment from within Validator method etc...  see https://github.com/akretion/sale-import/blob/14.0/sale_import_delivery_carrier/datamodels/delivery_carrier.py#L14 for exemple.

base_rest_demo/tests/test_new_api.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class TestPartner(NewServiceCommonCase):
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        cls.service = cls._get_service("partner")
+
+    def test_search(self):
+        self.env["res.partner"].create([{"name": f"P{idx}"} for idx in range(0, 1000)])
```
lmignon: Since the problem is not on the service layer, theses tests should be done into datamodels addon and should only involve the creation if a batch of model instances..

lmignon:
@etobella Ca n you explain how and why your code fix the original issue? I would like to understand if we can find a better solution to keep the property `env` into a datamodel instance

datamodel/core.py
```diff
@@ -378,17 +374,7 @@ def __init__(self, env, registry):
         self.registry = registry
 
     def __getitem__(self, key):
-        model = self.registry[key]
-        model.__init__ = functools.partialmethod(model.__init__, env=self.env)
-
-        @classmethod
-        def __get_schema_class__(cls, **kwargs):
-            cls = cls.__schema_class__(**kwargs)
-            cls._env = self.env
```
etobella, line 387: When the error was raised, this was being referenced every time, this showed that something wrong happened here.

Once found, the only option was trying to remove all dependencies on this.

etobella:
We can try to find a way to insert the env there, but the new definition must be always done from the registry IMO.

base_rest_demo/tests/test_new_api.py
etobella: Done

etobella:
I have been checking and I cannot find an easy way to add the env back. IMO we should proceed with this and we are trying to solve a major issue. If it is considered interesting to get back the `env` anyone interested can create a new PR

etobella:
> @etobella By removing the access to Environment from inside the Datamodel you remove the ability to access to the environment from within Validator method etc... see https://github.com/akretion/sale-import/blob/14.0/sale_import_delivery_carrier/datamodels/delivery_carrier.py#L14 for exemple.

I understand that we are loosing some functionality, but I think it is better to loose it than have something that raises errors

However, as a comment for future developers, it might be possible to add the env using `marshmallow.pre_load` (not sure, but it seems possible). Another option would be to use the context parameter from marsmallow. In any case, I am unable to dedicate more time to investigation.

lmignon:
@etobella I close this one since it's solved by #120 I'll do the forward port in 13 and 14. Thank you for all your investigations!

PR #120, opened 2021-02-23
[12.0] Fix  issue with datamodel, recursion limit

lmignon:
supersedes #118 includes #117


lmignon:
@etobella @sebastienbeau The issue is fixed without removing functionalities.  Thank you for all your investigations!

datamodel/core.py
```diff
@@ -388,6 +391,7 @@ def __get_schema_class__(cls, **kwargs):
             return cls
 
         model.__get_schema_class__ = __get_schema_class__
+        setattr(model, "__datamodel_init_patched", object())  # noqa: B010
```
sbidoul:
```suggestion
        setattr(model, "__datamodel_init_patched", True)  # noqa: B010
```

sebastienbeau:
@lmignon Thanks I will test it real soon !

datamodel/core.py
lmignon: Thank you!

lmignon:
@sebastienbeau I discovered an issue with the current implementation (and the previous one). The environment on the datamodel is not consistent.... IOW, You're not ensured that the property env on datamodel instance always returns the env from where the instance has been initialized..... I've to thing further how to fix this issue if it's possible.

etobella:
Not working properly, the environment is not correctly passed  to the datamodel.

Check the error using the following commit:
https://github.com/tegin/rest-framework/commit/4822c50efb575a2322059bf2cf370c234908ff83

---

I did not receive your comment on time, sorry

lmignon:
@etobella You're right, I just added the test on my side. The problem was already there into the previous implementation. I'll try to find a solution in the coming days...

rvalyi:
Hello I can confirm this fixed a real production issue I had on v12. Tested before and after, it really fixed it, thanks.

PR #126, opened 2021-03-09
[13.0][FIX] base_rest invalid dynamic object (controller) construction

acsonefho:
**Info:**
This PR fix 1 errors due to changes on new API:
1. Inheritance between controllers ignored (this PR);
2. `_default_auth` on controllers not applied on routes (I create another PR because it could be an expected behavior; lets discussion open: https://github.com/OCA/rest-framework/pull/130).

https://github.com/OCA/rest-framework/pull/48
So this bug in also in `12.0`, `13.0` and `14.0`.

**TODO:**
1. Unit test

### 1 - Controllers inheritance
**Bug - How to reproduce:**
Inherit an existing controller (to have a real case, I'm using as example the `shopinvader_locomotive` module who has a controller who uses `RestController`).
Try to overwrite a function (`_get_component_context(...)` for my example).
Currently if you put a break-point in `_get_component_context(...)` it's not loaded at all. As the controller is not really loaded (that's the issue).

```
from odoo.addons.shopinvader.controllers.main import InvaderController
from odoo.http import request


class LocomotiveInvaderController(InvaderController):
    @classmethod
    def _get_client_header(cls, headers):
        res = {}
        for key, val in headers.items():
            if key.upper().startswith("HTTP_INVADER_CLIENT_"):
                res[key.replace("HTTP_INVADER_CLIENT_", "")] = val
        return res

    def _get_component_context(self):
        res = super()._get_component_context()
        headers = request.httprequest.environ
        res["client_header"] = self._get_client_header(headers)
        return res
```

**Why the controller is not loaded?**
When Odoo starts, the following line checks if the `Controller` (`LocomotiveInvaderController`) that we're trying to load is a `RestController` (quite logic, the purpose is to load specific controllers for Rest only).
But as the `LocomotiveInvaderController` inherit the `InvaderController`, the `RestController` is not (directly) into the `bases`. So it's ignored.
https://github.com/acsone/rest-framework/blob/e572040ec41383d07a106742ceb185f274203391/base_rest/controllers/main.py#L39
But in this case, we don't want to inherit `RestController` because we want to inherit the existing behavior of the `InvaderController`.
So the line referenced just before should check if at least one "super-type" is a `RestController` to know if it must be considered as a Rest controller or not.

**Additionnal note:**
As there is a `metaclass` specified on `RestController`, it's not possible to use the `isinstance(...)` who'll return this `metaclass` (`RestControllerType`) and we need to have `RestController`.

### DIFF
--- a/base_rest/controllers/main.py
+++ b/base_rest/controllers/main.py
@@ -36,7 +36,9 @@ class RestControllerType(ControllerType):
             # our RestConrtroller must be a direct child of Controller
             bases += (Controller,)
         super(RestControllerType, cls).__init__(name, bases, attrs)
-        if "RestController" not in globals() or RestController not in bases:
+        if "RestController" not in globals() or not any(
+            issubclass(b, RestController) for b in bases
+        ):
             return
         # register the rest controller into the rest controllers registry
         root_path = getattr(cls, "_root_path", None)


acsonefho:
@lmignon can you take a look please?

base_rest/controllers/main.py
```diff
@@ -36,7 +36,9 @@ def __init__(cls, name, bases, attrs):  # noqa: B902
             # our RestConrtroller must be a direct child of Controller
             bases += (Controller,)
         super(RestControllerType, cls).__init__(name, bases, attrs)
-        if "RestController" not in globals() or RestController not in bases:
+        if "RestController" not in globals() or not any(
```
lmignon, line 39: :+1:

base_rest/models/rest_service_registration.py
```diff
@@ -160,10 +163,29 @@ def fix(self):
...
+        applied on every routes (cfr @route odoo's decorator).
+        This auth attribute should be applied only if the route doesn't already
+        define it.
+        :return:
+        """
+        # If the controller class doesn't have the _default_auth, we don't
+        # have to define it on every routes.
+        if not hasattr(self._controller_class, "_default_auth"):
+            return
+        for name, method in _inspect_methods(self._controller_class):
```
lmignon: @acsonefho I'm not sure if we should apply the default auth méthod. If someone add a route by hand, IMO it's its responsibility to specify the right auth on the route. What do you think about?

lmignon (commented):
Thank you for these fixes!  We should find a way to test this part of code to avoid future regressions...

base_rest/models/rest_service_registration.py
```diff
@@ -160,10 +163,29 @@ def fix(self):
...
                 continue
-            if not hasattr(method, "routing"):
+            if not hasattr(method, ROUTING_DEC):
                 methods_to_fix.append(method)
         for method in methods_to_fix:
             self._fix_method_decorator(method)
+        self._update_auth_method_controller()
+
+    def _update_auth_method_controller(self):
+        """
```
simahawk: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

base_rest/models/rest_service_registration.py
simahawk: OTOH it makes sense that if you specify it at controller level, it's applied too all endpoints.
And below, it checks if `auth` is specified, if yes, it's kept. No?

base_rest/models/rest_service_registration.py
```diff
@@ -29,6 +29,9 @@
 )
 from ..tools import _inspect_methods
 
+# Decorator attribute added on a route function (cfr Odoo's route)
+ROUTING_DEC = "routing"
```
simahawk:
```suggestion
ROUTING_DECORATOR_ATTR = "routing"
```

simahawk (commented):
thanks for fixing this!

base_rest/models/rest_service_registration.py
acsonefho: @simahawk Yes. If the `auth` is specified on the `@route(...)`, it's kept.

lmignon:
@acsonefho Could you check the travis error PLZ.

base_rest/models/rest_service_registration.py
acsonefho: I split this PR. So the fix on the route is now here: https://github.com/OCA/rest-framework/pull/130

lmignon (approved):
Thank you @acsonefho 
LGTM

PR #128, opened 2021-03-12
[12.0][FIX] base_rest: Allow disabling a service when _is_rest_service_component is False

marcelsavegnago:
Currently, only the presence of the attribute is inspected, which means

  _is_rest_service_component = False

Has the same effect as:

  _is_rest_service_component = True


marcelsavegnago:
ping @guewen

guewen (approved):

marcelsavegnago:
ping @OCA/rest-framework-maintainers @OCA/rest-framework-maintainers-psc-representative

PR #130, opened 2021-03-17
[13.0][FIX] base_rest: fix default_auth not automatically applied on routes

acsonefho:
Issue/regression (IMO) caused by New API: https://github.com/OCA/rest-framework/pull/48.
Previously, this fix was integrated into https://github.com/OCA/rest-framework/pull/126.

This bug (if it is) in also in `12.0`, `13.0` and `14.0`.

**TODO:**
1. Unit test

Previously (before changes to new API), `auth` on routes were automatically applied based on the `_default_auth` defined on the controller.

As the code has been removed due to new implementation, I add a new function to automatically apply this attribute.
Only if the route doesn't specify manually the `auth` attribute (who should still possible).

### DIFF
--- a/base_rest/models/rest_service_registration.py
+++ b/base_rest/models/rest_service_registration.py
@@ -29,6 +29,9 @@ from ..core import (
 )
 from ..tools import _inspect_methods
 
+# Decorator attribute added on a route function (cfr Odoo's route)
+ROUTING_DECORATOR_ATTR = "routing"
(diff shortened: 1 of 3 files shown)


lmignon (approved):
Thank you @acsonefho for this fix!

lmignon:
@acsonefho I have improved the fix and added the appropriate tests

PR #132, opened 2021-03-19
[13.0][FIX] base_rest: Includes default parameters and responses into the g…

### LINKED ISSUE #103: api-docs broken as headers are not documented with new implementation
Before the refactoring #71 every service was relying on `_get_openapi_default_parameters` to document endpoints (eg: required headers). 
When using `restapi.method` this is ignored and docs are not usable at all.

@lmignon shouldn't `RestApiMethodRoutePlugin` take care of this as well?

Example:

```
@restapi.method(
        [(["/", "/search"], "GET")], auth="api_key",
    )
    def search(self):
        return self._paginate_search()
```

IMO we should support old behavior and allow to pass per method parameters via decorator.

lmignon:
…enerated openapi doc

fixes #103

ping @simahawk

### DIFF
--- a/base_rest/apispec/rest_method_param_plugin.py
+++ b/base_rest/apispec/rest_method_param_plugin.py
@@ -31,30 +31,41 @@ class RestMethodParamPlugin(BasePlugin):
         if not operations:
             return
         for method, params in operations.items():
-            input_param = routing.get("input_param")
-            output_param = routing.get("output_param")
-            if input_param and isinstance(input_param, RestMethodParam):
-                parameters = params.get("parameters", [])
-                # add default paramters provided by the sevice
-                parameters.extend(self._default_parameters)
-                if method == "get":
-                    # get quey params from RequestMethodParam object
-                    parameters.extend(
-                        input_param.to_openapi_query_parameters(self._service)
-                    )
-                    # sort paramters to ease comparison into unittests
-                else:
-                    # get requestBody from RequestMethodParam object
-                    request_body = params.get("requestBody", {})
-                    request_body.update(
-                        input_param.to_openapi_requestbody(self._service)
-                    )
-                    params["requestBody"] = request_body
-                parameters.sort(key=lambda a: a["name"])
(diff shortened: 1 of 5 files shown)


base_rest/apispec/rest_method_param_plugin.py
```diff
@@ -31,30 +31,41 @@ def operation_helper(self, path=None, operations=None, **kwargs):
...
-                parameters.sort(key=lambda a: a["name"])
+            parameters = self._generate_pamareters(routing, method, params)
+            if parameters:
                 params["parameters"] = parameters
-            if output_param and isinstance(output_param, RestMethodParam):
-                responses = params.get("responses", {})
-                # get response from RequestMethodParam object
-                responses.update(self._default_responses.copy())
-                responses.update(output_param.to_openapi_responses(self._service))
+            responses = self._generate_response(routing, method, params)
```
simahawk:
```suggestion
            responses = self._generate_responses(routing, method, params)
```
or, rename var to `response`? Do you expect multiple items?

base_rest/tests/test_openapi_generator.py
```diff
@@ -150,3 +150,58 @@ def update_name(self, _id, _name):
...
+
+            def _get_openapi_default_responses(self):
+                responses = super()._get_openapi_default_responses().copy()
+                responses["999"] = "TEST"
+                return responses
+
+        self._build_services(self, PartnerService)
+        service = self._get_service_component(self, "partner")
+        openapi = service.to_openapi()
+        self.assertTrue(openapi)
```
simahawk: this assert seems superfluos

simahawk (approved):
:tada: merci!

base_rest/apispec/rest_method_param_plugin.py
lmignon: responses is a list... I'll rename

base_rest/tests/test_openapi_generator.py
lmignon: :+1:

PR #148, opened 2021-04-29
[IMP] base_rest: Allow to pass the url parameters as kwargs

etobella:
With this change, parameters from the url, will be used as kwargs. It will not change the behavior of existing installations
Without this change, if you have a url, with two parameters, sometimes the parameters where received on the wrong order.

So, for example, using "/<int:partner_id>/<string:demo_field>/get", sometimes, the first parameter where the partner_id and sometimes the demo_field.

@lmignon

### DIFF
--- a/base_rest/components/service.py
+++ b/base_rest/components/service.py
@@ -134,7 +134,7 @@ class BaseRestService(AbstractComponent):
             return result
         return output_param.to_response(self, result)
 
-    def dispatch(self, method_name, *args, params=None):
+    def dispatch(self, method_name, *args, params=None, **kwargs):
         """
         This method dispatch the call to the final method.
         Before the call parameters are processed by the
@@ -155,9 +155,11 @@ class BaseRestService(AbstractComponent):
         if isinstance(secure_params, dict):
             # for backward compatibility methods expecting json params
             # are declared as m(self, p1=None, p2=None) or m(self, **params)
-            res = method(*args, **secure_params)
+            final_params = kwargs.copy()
+            final_params.update(secure_params)
+            res = method(*args, **final_params)
         else:
-            res = method(*args, secure_params)
+            res = method(*args, secure_params, **kwargs)
         self._log_call(method, params, secure_params, res)
         return self._prepare_response(method, res)
 
--- a/base_rest/controllers/main.py
+++ b/base_rest/controllers/main.py
@@ -138,8 +138,8 @@ class RestController(Controller, metaclass=RestControllerType):
             raise BadRequest()
         return True
 
-    def _process_method(self, service_name, method_name, *args, params=None):
+    def _process_method(self, service_name, method_name, *args, params=None, **kwargs):
         self._validate_method_name(method_name)
         with self.service_component(service_name) as service:
-            result = service.dispatch(method_name, *args, params=params)
+            result = service.dispatch(method_name, *args, params=params, **kwargs)
             return self.make_response(result)
--- a/base_rest/models/rest_service_registration.py
+++ b/base_rest/models/rest_service_registration.py
@@ -329,7 +329,17 @@ class RestApiServiceControllerGenerator(object):
                 default_route = routes[0]
                 rule = Rule(default_route)
                 Map(rules=[rule])
-                if rule.arguments:
+                if rule.arguments and routing["endpoint_params_to_kwargs"]:
+                    method = METHOD_TMPL_WITH_KWARGS.format(
+                        method_name=method_name,
+                        service_name=self._service_name,
+                        service_method_name=name,
+                        args=", ".join("%s=False" % arg for arg in rule.arguments),
+                        args_value=", ".join(
+                            "{}={}".format(arg, arg) for arg in rule.arguments
+                        ),
+                    )
+                elif rule.arguments:
                     method = METHOD_TMPL_WITH_ARGS.format(
                         method_name=method_name,
                         service_name=self._service_name,
@@ -364,6 +374,15 @@ def {method_name}(self, **kwargs):
     )
 """
 
+METHOD_TMPL_WITH_KWARGS = """
+def {method_name}(self, {args}, **kwargs):
+    return self._process_method(
+        "{service_name}",
+        "{service_method_name}",
+        params=kwargs,
+        {args_value},
+    )
+"""
 
 METHOD_TMPL_WITH_ARGS = """
 def {method_name}(self, {args}, **kwargs):
--- a/base_rest/restapi.py
+++ b/base_rest/restapi.py
@@ -12,7 +12,13 @@ from .tools import cerberus_to_json
 
 
 def method(
-    routes, input_param=None, output_param=None, auth=None, cors=None, csrf=False
+    routes,
+    input_param=None,
+    output_param=None,
+    auth=None,
+    cors=None,
+    csrf=False,
+    endpoint_params_to_kwargs=False,
 ):
     """Decorator marking the decorated method as being a handler for
     REST requests. The method must be part of a component inheriting from
@@ -38,6 +44,8 @@ def method(
     :param cors: The Access-Control-Allow-Origin cors directive value.
     :param bool csrf: Whether CSRF protection should be enabled for the route.
                       Defaults to ``False``
+    :param bool endpoint_params_to_kwargs: Pass the query parameters as keyword
+                arguments, following the same logic than Werkzeug
 
     """
 
@@ -54,6 +62,7 @@ def method(
             "routes": _routes,
             "input_param": input_param,
             "output_param": output_param,
+            "endpoint_params_to_kwargs": endpoint_params_to_kwargs,
         }
 
         @functools.wraps(f)
--- a/base_rest_demo/services/partner_new_api_services.py
+++ b/base_rest_demo/services/partner_new_api_services.py
@@ -71,3 +71,13 @@ class PartnerNewApiService(Component):
 
     def _get(self, _id):
         return self.env["res.partner"].browse(_id)
+
+    @restapi.method(
+        [(["/<int:partner_id>/<string:demo_field>/get"], "GET")],
+        output_param=Datamodel("partner.info"),
+        auth="public",
+        endpoint_params_to_kwargs=True,
+    )
+    def get_with_kwargs(self, partner_id=False, demo_field=False):
+        assert demo_field
+        return self.get(partner_id)


base_rest/models/rest_service_registration.py
```diff
@@ -329,7 +329,14 @@ def _generate_methods(self):
                 default_route = routes[0]
                 rule = Rule(default_route)
                 Map(rules=[rule])
-                if rule.arguments:
+                if rule.arguments and routing["params_as_kwargs"]:
+                    method = METHOD_TMPL_WITH_KWARGS.format(
```
lmignon, line 333: Thank you @etobella. 
 As stated by the werkzeug documentation `besides the rule-string all arguments *must* be keyword arguments  in order to not break the application on Werkzeug upgrades` https://github.com/pallets/werkzeug/blob/master/src/werkzeug/routing.py#L548

The problem here is to be able to dissociate path parameters (validated by werkzeuk) from others parameters (query / json) validated by the rest framework. In the current implementation I expect that path parameters are always pass to the final method into the same order as the one defined into the url. Unfortunately it's seems it's not always the case when werkzeug calls the generated method.

IMO it would be safe to be able to not break the convention that path parameters are *args in the same order that into the path definition and **params are query/json paramters. We could maybe achieve this by only modifying the generated method....


```python
METHOD_TMPL_WITH_ARGS = """
def {method_name}(self, {path_parameters}, **kwargs):
    return self._process_method(
        "{service_name}",
        "{service_method_name}",
        *[{args}],
        params=kwargs
    )
"""

if rule.arguments:
    method = METHOD_TMPL_WITH_ARGS.format(
        method_name=method_name,
        service_name=self._service_name,
        service_method_name=name,   
        path_parameters = ", " . join(["%s=None" for arg in rule.arguments]),
        args=", ".join(rule.arguments),
    )
    

```

What do you think about?

etobella: The problem I found is that in some cases (on my case, if the a rest call is sent before the modules are fully loaded), werkzeug is passing the parameters in the inverse order, so, when generating the method, the passed parameters will not preserve the order.
For example, If I have the
`/<string:param1>/<int:param2>`
the order is important, and it might happen two options:
```
def my_method(self, param1=None, param2=None, **kwargs):
    return self._process_method(
        "my_method",
        "{service_method_name}",
        *[param1, param2],
        params=kwargs
    )
```
or
```
def my_method(self, param2=None, param1=None, **kwargs):
    return self._process_method(
        "my_method",
        "{service_method_name}",
        *[param2, param1],
        params=kwargs
    )
```
As you can see, the results will be different

etobella: Also, I knew the changes might be a controversial, for that reason, the option is only added if you really need it.

lmignon: @etobella  In order to preserve the order in every cases we could extract the parameters list into the right order from `rule.string` https://github.com/pallets/werkzeug/blob/master/src/werkzeug/routing.py#L551

etobella: I think that making this is even more complex, as we are trying to fix something because we are not using the expected behavior

> As stated by the werkzeug documentation besides the rule-string all arguments *must* be keyword arguments in order to not break the application on Werkzeug upgrades https://github.com/pallets/werkzeug/blob/master/src/werkzeug/routing.py#L548

base_rest/components/service.py
```diff
@@ -155,9 +155,11 @@ def dispatch(self, method_name, *args, params=None):
         if isinstance(secure_params, dict):
             # for backward compatibility methods expecting json params
             # are declared as m(self, p1=None, p2=None) or m(self, **params)
-            res = method(*args, **secure_params)
+            final_params = kwargs.copy()
```
etobella, line 158: The params from the query takes precedence from the kwargs

base_rest/controllers/main.py
```diff
@@ -138,8 +138,8 @@ def _validate_method_name(self, method_name):
             raise BadRequest()
         return True
 
-    def _process_method(self, service_name, method_name, *args, params=None):
+    def _process_method(self, service_name, method_name, *args, params=None, **kwargs):
```
lmignon, line 141: This signature change will break a lot of projets....

etobella: Why? I don't know a lot of projects with rest-framework, but no changes where required to make it work with mines and checking shopinvader this function is not changed

If you are talking about dispatch, this is true, as it is modified, but you will only need to make the change if you selected the params_as_kwargs, as the kwargs will be empty, so it should be no issue

lmignon:
@simahawk Your insight would be welcome here ... I made a mistake into the implementation when adding the support for path parameters... I would find a way to fix this issue without breaking existing installation and providing a clear rule on how parameters (path / query / json) are passed to methods.

base_rest/controllers/main.py
lmignon: @etobella `params_as_kwargs' should not be an option since we have a real issue here. That's why I try to find a solution to solve the issue without introducing a new parameter into the route definition nor modifying the methods signatures.

etobella: Solution is not simple, probably not possible on current installations.

With the current app, you could do the following:

```

    @restapi.method(
        [(["/get/<int:id", "GET")],
    )
    def get_data(self, _id):
```

and worked properly, you can see that the parameters have not the same name, as it depended on order, my proposal would be to set the params_as_kwargs as False and fix the issue on 15.0

etobella: For example, even on the tests of rest-framework, exists a case when the parameter name has been changed:

https://github.com/OCA/rest-framework/blob/12.0/base_rest/tests/test_openapi_generator.py#L32

Also, the automatic generated calls, like `get` have the same issue as we can see here:
https://github.com/OCA/rest-framework/blob/13.0/base_rest/models/rest_service_registration.py#L236
https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader/services/address.py#L31
We are expecting the field `_id` but on the url we are expecting `id`

My point is, we cannot use a dictionary as it can break current installations (the name used might be different) but  we should use key word arguments. Once we have arrived to this contradiction, there is not many options.

IMO, my proposal is similar to queue_job new implementation, when the old behavior had to be maintained, but the new logic will be the only one present after 15.0.

Another option is to show a message when it not fulfill the new behavior, this way, people have time to solve the issue on their installations

lmignon: Thank you @etobella for all this hard work. I'll take time to dig into this problem on mys side too. (Just to convince me that there is no other solution :smirk).

> Also, the automatic generated calls, like get have the same issue as we can see here:
> https://github.com/OCA/rest-framework/blob/13.0/base_rest/models/rest_service_registration.py#L236
> https://github.com/shopinvader/odoo-shopinvader/blob/14.0/shopinvader/services/address.py#L31

The generated routes for instance methods could be fixed to declare the path parameter as `_id` since it's the one  expected by the old api.

etobella: I know we can change them, I was trying to show, that this problem could be generic :sob:

base_rest/restapi.py
```diff
@@ -54,6 +60,7 @@ def decorator(f):
             "routes": _routes,
             "input_param": input_param,
             "output_param": output_param,
+            "params_as_kwargs": params_as_kwargs,
```
simahawk: docstring update missing

simahawk (commented):
It's hard to get a full picture of possible consequences and of possible better refactoring. 
From a quick look I think that we can live for a while w/ a specific param to turn on this feature especially if it does not break existing services when turned off.

BTW whatever solution you come up with, renaming those variables to something more meaningful would help (eg: `params` -> `request_params` or `endpoint_params`? ) as `params` and `kwargs` are pretty much the same in a generic world.

base_rest/restapi.py
etobella: Ok, I could use: `endpoint_params_to_kwargs`, WDYT?

etobella:
@simahawk I changed the name with the suggested name. I think now it is ready for review :wink:

etobella:
@lmignon can this be merged?

lmignon:
@etobella I'm still not convinced and have to dig into into this problem.  Moreover there is one thing that appeals to me. In a comment, you explain that this problem occurs when calling a service if the module is not completely initialized.

> The problem I found is that in some cases (on my case, if the a rest call is sent before the modules are fully loaded), 

This behavior should not be possible. The result of such a call is not predictable since in such circumstances the server is not into a consistent state. We should prevent the calls to the API while all the modules are not fully loaded. 

I don't want to block your PR, I just want to be sure to make the right fix and avoid to introduce different behavior for the same use case. It's a matter of consistency and maintainability.

etobella:
I understand your concern, but it is an issue. Also, about my point of the fully loaded. It happened to me that if you restarted the service (no updating) and a call was made during the process of update, the issue was raised and the system didn't solve it for the next calls. 

I cannot prevent the API calls while the modules are not fully loaded (as odoo is not consistent, but the calls are accepted but left on a waiting state), at least I don't know how to make it. Also, I made this double behavior as you could make a service with different names on the parameter between the defined rest and the function and I didn't want to break existing services. 

I think we should use my system for future versions as it is more consistent.

lmignon:
@etobella Ok to merge it in 13 and 14 but we rest_log should also be fixed. For 15 we must get rid of `*args`

@simahawk Do you agree?

simahawk:
Guys, thinking a bit more on this (and other details) I think we are looking at the problem from the wrong angle.
IMO we should drop all the arguments and pass everything via work ctx. This way, the real method never accepts args and on the work ctx we can have `original_params` (bare unsecure args) and `params` (secure args).
Thoughts?

lmignon:
My thoughts is that arguments must be explicitly passed to the method (no magic) for readability and maybe in the future to allow code type checking.. My initial mistake was that I did not conceptualize my reasoning enough before starting the implementation. On a conceptual point of view HTTP calls can provide arguments in 3 different ways. 

* Path parameters ( POST, GET, PUT, DELETE)
* Query parameters (GET)
* Body parameters (POST, PUT)

Since it should not be allowed to mix Query and Body parameters, we should define the following convention:

* Path parameters should always be passed as kwargs
* Query/Body parameters should always be passed as 'param' parameter

With such a convention, it becomes easy to explain how to define the signature of a method.

We must pay attention to first conceptualize our thinking before trying to put in place a technical solution. In the same spirit, 

I've been thinking all weekend about the relevance of using 'Component' in the current approach. I'm getting the feeling more and more that we could greatly simplify the approach and make things more explicit by doing without 'Component'. All we need is a way to allow extension of our services and runtime context compositions. All this should be drafted before moving forward but IMO, It is important to take a step back from the current approach by questioning the initial choices in relation to current and future needs. (An other story...)

simahawk:
hmm IMO readability is ok anyway.
Moreover, it's probably better/cleaner w/out params as params are defined in validators too.
Regarding type checking: IMO is useless in this ctx as we have validators that validate and clean them beforehand.
It also grant that methods won't be broken in the future if we change our minds regarding signatures.

On the conceptual part this is my POV: we have components that work on specific context. 
This context is given when the component is initialized and the request and all its params are (should be) part of it IMO.
In this sense, we are even misusing it by defining `@property` instead of overriding `__init__` (as per my CR last time).

Now, regarding using components: I think that's fine.
What is not fine - as per our recent discussion - is the complicated handling of the generated routes (of course there are good reasons that led us here). In this regard, I encourage you have a look at my work here https://github.com/OCA/edi/pull/450.
But this is another story :wink:

sbidoul:
> pass everything via work ctx

I'm not convinced with that approach. I think it would make the code harder to follow.

> Path parameters ( POST, GET, PUT, DELETE), Query parameters (GET), Body parameters (POST, PUT)

@lmignon FWIW, FastAPI allows [combining](https://fastapi.tiangolo.com/tutorial/body-multiple-params/) path, query and body, irrespective of the HTTP method. And I have used POST with query parameters myself.

etobella:
@lmignon should we reopen it and merge? Or what can we do with this?

lmignon:
@etobella can you try the fix here https://github.com/OCA/rest-framework/pull/258

PR #158, opened 2021-05-18
rest_log: drop post init hook for shopfloor.log

simahawk:
This hook is not needed anymore.
If you were using shopfloor before this module was introduced
you already migrated data.


jgrandguillaume (approved):
Thx

PR #159, opened 2021-05-20
[13.0][FIX] datamodel: Error when NestedModel inside other field (Example: field.List())

CarlosRoca13:
Before this patch, when we had a NestedModel inside a field.List the datamodel could not be constructed because the List object hasn't the variable _env.

Now we get the absolute parent to avoid this error and allow to make list of NestedModels.

Capture before the patch:
image

Capture after the patch:
image

cc @Tecnativa TT27860

### DIFF
--- a/datamodel/fields.py
+++ b/datamodel/fields.py
@@ -30,10 +30,18 @@ class NestedModel(Nested):
     @property
     def schema(self):
         if not self.nested:
-            self.nested = self.parent._env.datamodels[
+            # Get the major parent to avoid error of _env does not exist
+            super_parent = None
+            parent = self
+            while not super_parent:
+                if not hasattr(parent, "parent"):
+                    super_parent = parent
+                    break
+                parent = parent.parent
+            self.nested = super_parent._env.datamodels[
                 self.datamodel_name
             ].__schema_class__
-            self.nested._env = self.parent._env
+            self.nested._env = super_parent._env
         return super(NestedModel, self).schema
 
     def _deserialize(self, value, attr, data, **kwargs):
--- a/datamodel/static/description/index.html
+++ b/datamodel/static/description/index.html
@@ -480,13 +480,21 @@ If you spotted it first, help us smashing it by providing a detailed and welcome
 </ul>
 </div>
 <div class="section" id="contributors">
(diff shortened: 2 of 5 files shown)


CarlosRoca13:
Please @lmignon, can you check it?

fd-oncodna:
While I do agree that it is a bug and it should be fixed, nesting multiple objects should not be achieved by combining a `NestedModel` within a `List` field, but rather by declaring the nested model as `many=True` (`images = NestedModel('my.image.model', many=True)`)

datamodel/fields.py
```diff
@@ -30,10 +30,18 @@ def __init__(self, nested, **kwargs):
     @property
     def schema(self):
         if not self.nested:
-            self.nested = self.parent._env.datamodels[
+            # Get the major parent to avoid error of _env does not exist
+            super_parent = None
+            parent = self
+            while not super_parent:
+                if "parent" not in dir(parent):
```
lmignon: if not hasattr(parent, "parent"):

lmignon (changes requested):
Thank you @CarlosRoca13.
A little comment and can you split your changes into 2 commits? 1 for the fix and another one for the new readme. 

Regards

CarlosRoca13:
@lmignon changes done!

datamodel/static/description/index.html
```diff
@@ -3,7 +3,11 @@
 <html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
+<<<<<<< HEAD
```
lmignon: Remaining conflict info lines

lmignon (changes requested):

lmignon:
@CarlosRoca13 Take care of the @fd-oncodna advice when you design your API!
Moreover, can you add a testcase for your fix plz

regards.

CarlosRoca13:
@lmignon @fd-oncodna Thanks, I will keep it on mind :)

Changes done!

lmignon (approved):
Thank you @CarlosRoca13 LGTM!

pedrobaeza:
Can this be merged?

pedrobaeza:
Thanks!

Issue #164, opened 2021-06-09
Licensing of datamodel modules


davejames:
The rest of the modules in the rest-framework repo are all licensed under LGPL since https://github.com/OCA/rest-framework/pull/28 with the exception of the `datamodel` and `base_rest_datamodel` which remain licensed under AGPL.

Is there any possibility to relicense these modules for the same rationale as https://github.com/OCA/rest-framework/pull/28? I am happy to open a PR if there is no objection.

davejames:
Sorry for the tag, but @lmignon would you mind considering this? I think you're the author of these modules as well as the re-licensing PR I quoted.

lmignon:
@davejames It's OK for me relicence to LGPL. Feel free to make a PR in this way...

davejames:
Thanks @lmignon, I have created PRs for 12.0, 13.0 and 14.0

simahawk:
all merged

PR #28, opened 2019-06-17
[CHG] base_rest: Use more permissive licence: AGPL-> LGPL

lmignon:
This change will allow to use this framwork addon to build services on top of OE and LGPL addons.

@sebastienbeau @bealdav Do you agree?

PR #166, opened 2021-06-18
[12.0][CHG] datamodel: Use more permissive licence: AGPL-> LGPL

davejames:
The rest of the modules in the rest-framework repo are all licensed under LGPL since https://github.com/OCA/rest-framework/pull/28 with the exception of the `datamodel` and `base_rest_datamodel` which remain licensed under AGPL.

This PR relicenses these modules to LGPL-3. Closes https://github.com/OCA/rest-framework/issues/164


simahawk (approved):
fine for me. @lmignon ping :)

lmignon (approved):

PR #170, opened 2021-06-20
[13.0][IMP] base_rest_datamodel: allow to receive `partial` datamodels as params

fdegrave:
In particular, this feature allows one to use datamodels to make partial updates through `PUT` requests. In this case, the endpoint declaration would look like this:

```python
    @restapi.method(
        [(["/"], "PUT")],
        input_param=Datamodel("partner.info", partial=True),
        output_param=Datamodel("partner.info"),
    )
```

If `partial` is not set (or without this feature), the params sent by the client in the request must contain values for all the fields declared as `required` in the datamodel. Note that it uses an existing functionality of `marshmallow_objects`, as a call to `dump` of a `Datamodel` declared as `partial` would contain only the subset of fields provided in the params.

### DIFF
--- a/base_rest_datamodel/restapi.py
+++ b/base_rest_datamodel/restapi.py
@@ -12,21 +12,27 @@ from odoo.addons.base_rest import restapi
 
 
 class Datamodel(restapi.RestMethodParam):
-    def __init__(self, name, is_list=False):
+    def __init__(self, name, is_list=False, partial=None):
         """
 
-        :param name: The datamdel name
+        :param name: The datamodel name
         :param is_list: Should be set to True if params is a collection so that
                         the object will be de/serialized from/to a list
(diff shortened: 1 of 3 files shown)


gdgellatly:
+1 with forward/backports. As far as I have observed, your description is not quite correct. It only must contain all the fields declared as required. But that is v14, so maybe 13 is different.

fdegrave:
No you are right, even in v13; I updated the description.

lmignon:
@fdegrave Thank you for this enhancement. Can you add a unitest for this change? Once added I'll merge it

lmignon (approved):

lmignon:
@fdegrave Can you rebase and fix the conflicts? Once done I'll merge your PR.

PR #171, opened 2021-06-23
[13.0][ADD] rest_authenticate: provide auth endpoints

fdegrave:
This `rest_authenticate` module adds API endpoints to deal with session authentication and logout.

Usage is:

## Authentication

To authenticate you need to `POST` a request on `[ODOO HOST]/session_rest/authenticate` with the following body:

```json
    {
        "db": [DB_NAME],
        "login": [LOGIN],
        "password": [PASSWORD]
    }
```

If the authentication is successful, the response will contain (in addition to the usual response of the JSON-RPC authentication):

```json
    {
        ...
        "session": {
            "sid": "ff6b4bac7a590e7960abfc0ac38361433ecac1d6",
            "expires_at": "2021-09-21 16:53:56"
        }
    }
```

This `sid` value can then be used in the header of the subsequent requests under the `X-Openerp-Session-Id` key.

## Logout

To logout you need to `POST` a request on `[ODOO HOST]/session_rest/logout` with an empty body.


fdegrave:
Of course this is just a proposal; I use it for my own needs and I see some others would benefit from it too. It is basically the exact same authentication scheme (and code) as the classical JSON/XML-RPC but it allows to avoid JSON/XML-RPC altogether.

rest_authenticate/services/session.py
```diff
@@ -0,0 +1,55 @@
...
+        httprequest.session.modified = True
+
+
+class SessionAuthenticateService(Component):
+    _inherit = "base.rest.service"
+    _name = "session.authenticate.service"
+    _usage = "authenticate"
+    _collection = "session.rest.services"
+
+    @restapi.method([(["/"], "POST")])
```
simahawk: shouldn't be `/authenticate`?

rest_authenticate/services/session.py
```diff
@@ -0,0 +1,55 @@
...
+        return result
+
+
+class SessionLogoutService(Component):
+    _inherit = "base.rest.service"
+    _name = "session.logout.service"
+    _usage = "logout"
+    _collection = "session.rest.services"
+
+    @restapi.method([(["/"], "POST")])
```
simahawk: shouldn't be `/logout`?

rest_authenticate/services/session.py
```diff
@@ -0,0 +1,55 @@
...
+    _name = "session.authenticate.service"
+    _usage = "authenticate"
+    _collection = "session.rest.services"
+
+    @restapi.method([(["/"], "POST")])
+    def authenticate(self):
+        params = request.params
+        request.session.authenticate(params["db"], params["login"], params["password"])
+        result = request.env["ir.http"].session_info()
+        if request.session.rotate:
```
simahawk: this check is done already in `_rotate_session`

rest_authenticate/services/session.py
```diff
@@ -0,0 +1,55 @@
...
+            "expires_at": fields.Datetime.to_string(expiration),
+        }
+        return result
+
+
+class SessionLogoutService(Component):
+    _inherit = "base.rest.service"
+    _name = "session.logout.service"
+    _usage = "logout"
+    _collection = "session.rest.services"
```
simahawk: maybe a base component to inherit from?

simahawk (changes requested):
thanks for your proposal!

Would you mind adding tests?

rest_authenticate/services/session.py
fdegrave: No, because `_usage` is already at the end of the url.

rest_authenticate/services/session.py
fdegrave: Indeed, I remove it.

rest_authenticate/services/session.py
fdegrave: Probably; I first thought of doing everything in the same class, but that means that I will end up with a uselessly long url: `[controller route]/[_usage]/[service route]`, something like `session_rest/session/authenticate` in my case. Otherwise I can do it with the "old style" API, but I figured "new is always better" :-)

rest_authenticate/services/session.py
fdegrave: Same as previous comment.

fdegrave:
> Would you mind adding tests?

Tests added!

rest_authenticate/services/session.py
simahawk: yeah, I know, but it seems weird to have only the root path there.

@lmignon what would be the recommended way in this case?
I'd say that if you specify the route, you should take full control of it. It would also make clear which endpoint you are defining.

rest_authenticate/services/session.py
simahawk: I'd change the root path to `/session/` and have only `_usage = "auth"` so that you'll have something like `/session/auth/login|logout`

fdegrave:
Is there an issue with Travis? I get strange errors in the tests, they seem independent from my code -- I get the same error in another PR after removing a single line of comment (used to pass the tests fine before).

rest_authenticate/services/session.py
fdegrave: Ok good idea.

lmignon: :+1: for @simahawk proposal

fdegrave:
There, I made the modifications as discussed. But I still get this strange errors during validation in Travis: 

```
File "/home/travis/odoo-13.0/odoo/modules/module.py", line 609, in run
test(result)
File "/opt/python/3.6.7/lib/python3.6/unittest/suite.py", line 84, in __call__
return self.run(*args, **kwds)
File "/opt/python/3.6.7/lib/python3.6/unittest/suite.py", line 112, in run
self._tearDownPreviousClass(test, result)
File "/home/travis/odoo-13.0/odoo/tests/common.py", line 251, in _tearDownPreviousClass
previousClass.doClassCleanups()

AttributeError: type object 'TestBuildDatamodel' has no attribute 'doClassCleanups'
```

Any insight on the reason it happens? I get the exact same errors on another (unrelated) branch, so it has nothing to do with my PR I think.

simahawk:
> There, I made the modifications as discussed. But I still get this strange errors during validation in Travis:
> 
> ```
> File "/home/travis/odoo-13.0/odoo/modules/module.py", line 609, in run
> test(result)
> File "/opt/python/3.6.7/lib/python3.6/unittest/suite.py", line 84, in __call__
> return self.run(*args, **kwds)
> File "/opt/python/3.6.7/lib/python3.6/unittest/suite.py", line 112, in run
> self._tearDownPreviousClass(test, result)
> File "/home/travis/odoo-13.0/odoo/tests/common.py", line 251, in _tearDownPreviousClass
> previousClass.doClassCleanups()
> 
> AttributeError: type object 'TestBuildDatamodel' has no attribute 'doClassCleanups'
> ```
> 
> Any insight on the reason it happens? I get the exact same errors on another (unrelated) branch, so it has nothing to do with my PR I think.

please check how I fixed it on `component_event` https://github.com/OCA/connector/commit/0d348f7d72b4e232d7f5954085d49f3e7d7cec21

fdegrave:
Ok thanks! However I think it won't fix things here, because my tests are not simple `unittest.TestCase`, it must come from another module in this repo (like [this one](https://github.com/OCA/rest-framework/blob/1a21325204dc15aaf5dde7b1e8fc599eb313994c/base_rest/tests/test_cerberus_validator.py#L15)).

I think this PR in not the place to fix tests outside of `rest_authenticate`...

lmignon:
> Ok thanks! However I think it won't fix things here, because my tests are not simple `unittest.TestCase`, it must come from another module in this repo (like [this one](https://github.com/OCA/rest-framework/blob/1a21325204dc15aaf5dde7b1e8fc599eb313994c/base_rest/tests/test_cerberus_validator.py#L15)).
> 
> I think this PR in not the place to fix tests outside of `rest_authenticate`...

It' not possible to merge PR if tests fails.

2 possibilities:

* Tests are fixed into this PR in a specific commit
* Tests are fixed into an other PR and you rebase your branch on the other one (my preference)

Thank you for all your work

Regards

fdegrave:
Here you go: https://github.com/OCA/rest-framework/pull/176

fdegrave:
Great, everything works in https://github.com/OCA/rest-framework/pull/176 ! I guess I have to wait for you to merge it before I can rebase this one?

fdegrave:
Rebase is done. I'll squash my commits.

fdegrave:
@simahawk : I made all changes requested but since I squashed my commits and force-pushed the change request disappeared...

rest_authenticate/readme/USAGE.rst
```diff
@@ -0,0 +1,29 @@
...
+
+    {
+        ...
+        "session": {
+            "sid": "ff6b4bac7a590e7960abfc0ac38361433ecac1d6",
+            "expires_at": "2021-09-21 16:53:56"
+        }
+    }
+
+This :code:`sid` value can then be used in the header of the subsequent requests under the :code:`X-Openerp-Session-Id` key.
```
simahawk:
```suggestion
This :code:`sid` value can then be sent in subsequent requests in the following ways:

* header :code:`X-Openerp-Session-Id`
* cookie named `session_id`
* request param `session_id`
```
am I wrong?

rest_authenticate/readme/USAGE.rst
```diff
@@ -0,0 +1,29 @@
+Authentication
+~~~~~~~~~~~~~~
+
+To authenticate you need to :code:`POST` a request on :code:`[ODOO HOST]/session/auth/login` with the
+following body::
+
+    {
+        "db": [DB_NAME],
```
simahawk: Is the `db` really mandatory? AFAIK it can be automatically determined. Check for `db_monodb`.

rest_authenticate/tests/test_rest_session.py
```diff
@@ -0,0 +1,50 @@
...
+import odoo.tests.common
+from odoo.http import root
+from odoo.tests import new_test_user
+from odoo.tests.common import HttpCase
+
+HOST = "127.0.0.1"
+PORT = odoo.tools.config["http_port"]
+
+
+@odoo.tests.common.at_install(False)
```
simahawk: FYI you can now `@tagged` to set this flags

rest_authenticate/tests/test_rest_session.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @odoo.tools.mute_logger("odoo.addons.base_rest.http")
+    def test_01(self):
+        """Test authentication"""
+        params = {
+            "db": self.env.cr.dbname,
+            "login": self.user.login,
+            "password": self.password,
+        }
+        url = "http://{}:{}{}".format(HOST, PORT, "/session/auth/login")
```
simahawk: you don't need to compute the URL yourself, just pass the path

rest_authenticate/tests/test_rest_session.py
```diff
@@ -0,0 +1,50 @@
...
+        response = self.opener.post(url, json=params)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.headers["Content-Type"], "application/json")
+        data = response.json()
+        session = root.session_store.get(data["session"]["sid"])
+        self.assertIsNotNone(session)
```
simahawk:
```suggestion
```
superfluous as you check below that you got something in login, no?

rest_authenticate/readme/USAGE.rst
fdegrave: No, you are right.

rest_authenticate/tests/test_rest_session.py
fdegrave: Great! Thanks, much shorter

rest_authenticate/tests/test_rest_session.py
fdegrave: That's only true if I call `url_open`, but in that case I cannot control the http method to use -- no `POST` with an empty body for example.

simahawk: oh, you are right. Never mind then :)

rest_authenticate/services/session.py
```diff
@@ -0,0 +1,48 @@
...
+        _rotate_session(request)
+        request.session.rotate = False
+        expiration = datetime.datetime.utcnow() + datetime.timedelta(days=90)
+        result["session"] = {
+            "sid": request.session.sid,
+            "expires_at": fields.Datetime.to_string(expiration),
+        }
+        return result
+
+    @restapi.method([(["/logout"], "POST")])
```
lmignon:
```suggestion
    @restapi.method([(["/logout"], "POST")], auth="user")
```

rest_authenticate/services/session.py
```diff
@@ -0,0 +1,48 @@
...
+        httprequest.session.modified = True
+
+
+class SessionAuthenticationService(Component):
+    _inherit = "base.rest.service"
+    _name = "session.authenticate.service"
+    _usage = "auth"
+    _collection = "session.rest.services"
+
+    @restapi.method([(["/login"], "POST")])
```
lmignon:
```suggestion
    @restapi.method([(["/login"], "POST")], auth="public")
```

rest_authenticate/controllers/main.py
```diff
@@ -0,0 +1,7 @@
+from odoo.addons.base_rest.controllers import main
+
+
+class SessionRestController(main.RestController):
+    _root_path = "/session/"
+    _collection_name = "session.rest.services"
+    _default_auth = "public"
```
lmignon:
```suggestion

```

lmignon:
@fdegrave Great job! I improved base_rest into 14.0 #180 Thanks to these improvements it's now possible to complete the openapi documentation about the authentication applied to each service method. see #182 and #181 . We also now include into the documentation the definition for the `auth="user"` https://github.com/OCA/rest-framework/blob/14.0/base_rest/apispec/rest_method_security_plugin.py#L20
In order to to kepp the same naming convention for addon managing the authentication in base_rest it would be nice to rename your addon as `base_rest_auth_user_service" What do you thing about?

rest_authenticate/readme/USAGE.rst
```diff
@@ -0,0 +1,34 @@
...
+            "sid": "ff6b4bac7a590e7960abfc0ac38361433ecac1d6",
+            "expires_at": "2021-09-21 16:53:56"
+        }
+    }
+
+This :code:`sid` value can then be sent in subsequent requests in the following ways:
+
+* header :code:`X-Openerp-Session-Id`
+* cookie named `session_id`
+* request param `session_id`
```
sbidoul: Out of curiosity, what are the use cases where the session cookie cannot be used and `X-Openerp-Session-Id` or the request param are necessary ?

fdegrave: Good question! I personally don't do this, but I've already seen people blocking cookies altogether when making http requests -- for some reason. I guess it is not used much since the `X-Openerp-Session-Id` key has never been updated to something more odoo-ish.

fdegrave: Maybe there is another, better reason I'm not aware of. I'm curious too!

fdegrave:
> it would be nice to rename your addon as `base_rest_auth_user_service" What do you thing about?

Sure, sounds fine to me. I'll do that.

lmignon (approved):
LGTM (Code review)

lmignon:
@simahawk Can you update your review?

simahawk (approved):

PR #174, opened 2021-06-29
[14.0] base_rest multiple http methods and cors OPTIONS

sbidoul:
- allow multiple http methods when declaring a route
- automatically enable then OPTIONS method when cors is set (otherwise the Odoo request handler rejects it with a 405, which is not good since OPTIONS must be handled correctly when doing cross origin calls in a browser app)

### DIFF
--- a/base_rest/restapi.py
+++ b/base_rest/restapi.py
@@ -43,10 +43,13 @@ def method(
 
     def decorator(f):
         _routes = []
-        for paths, http_method in routes:
+        for paths, http_methods in routes:
             if not isinstance(paths, list):
                 paths = [paths]
-            _routes.append(([p for p in paths], http_method))
+            if not isinstance(http_methods, list):
+                http_methods = [http_methods]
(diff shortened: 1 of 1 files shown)


lmignon (approved):

simahawk (approved):
LG

base_rest/restapi.py
```diff
@@ -35,18 +35,25 @@ def method(
...
         _routes = []
-        for paths, http_method in routes:
+        for paths, http_methods in routes:
             if not isinstance(paths, list):
                 paths = [paths]
-            _routes.append(([p for p in paths], http_method))
+            if not isinstance(http_methods, list):
+                http_methods = [http_methods]
+            if cors and "OPTIONS" not in http_methods:
+                http_methods.append("OPTIONS")
```
sbidoul, line 54: hm, this should better be
```suggestion
                http_methods = http_methods + ["OPTIONS"]
```

PR #177, opened 2021-07-02
[13.0][ADD] model_serializer: module to facilitate the (de)serialization of Odoo Models into DataModels

fdegrave:
This model adds functionalities similar to (a subset of) what offers Django Rest Framework through their Model Serializers (hence the name).

The `ModelSerializer` class inherits from the `Datamodel` class and adds functionalities. Therefore any class inheriting from `ModelSerializer` can be used the exact same way as any other `Datamodel`.

Basic example:

```python
    from odoo.addons.model_serializer.core import ModelSerializer

    class PartnerInfo(ModelSerializer):
        _name = "partner.info"
        _model = "res.partner"
        _model_fields = ["id", "name", "country_id"]
```

This `ModelSerializer` also provides methods to easily transform a recordset into a corresponding datamodel, and vice-versa. Please read the (still very short) doc to have more info on how it works.

~~I know there are no tests yet, but I will add them in case that the module is accepted.~~ Test added! Not full yet, but quite complete


lmignon:
same as in #193

PR #180, opened 2021-07-19
[14.0] base_rest: improve service context and security

lmignon:
Improvements:

* This PR introduces a new component `BaseRestServiceContextProvider`to be used to extend the component context that will be provided to the service layer when a method is called. Prior to this change, the only way to extend the context was to extend the default method `_get_component_context` provided by the RestController class. Since only one RestController should be registered into the database for a root_path and collection name, this mechanism give you the opportunity to extend the context provided by a base addon into specialized addons
*  If 2 `RestController`s are registered for the same root_path and  collection name, an error is now printed into the log at server startup
* It's now possible to output into the openapi doc, the authentication mechanism required by a method.
* The management of defaults route attributes defined on the RestController has been improved (applied only if not provided by a `@http.route` nor `|restapi.method`decorators)  
* Add a new kind of restapi method parameter `BinaryData`to be used to decalre method requesting/returning files

### DIFF
--- a/base_rest/controllers/main.py
+++ b/base_rest/controllers/main.py
@@ -44,8 +44,7 @@ class RestControllerType(ControllerType):
         root_path = getattr(cls, "_root_path", None)
         collection_name = getattr(cls, "_collection_name", None)
         if root_path and collection_name:
-            if not hasattr(cls, "_module"):
-                cls._module = _get_addon_name(cls.__module__)
+            cls._module = _get_addon_name(cls.__module__)
             _rest_controllers_per_module[cls._module].append(
                 {
                     "root_path": root_path,
@@ -79,13 +78,23 @@ class RestController(Controller, metaclass=RestControllerType):
     # Whether CSRF protection should be enabled for the route.
     _csrf = False
 
+    _component_context_provider = "component_context_provider"
+
     def _get_component_context(self):
         """
         This method can be inherited to add parameter into the component
         context
         :return: dict of key value.
         """
-        return {"request": request}
+        collection = self.collection
+        work = WorkContext(
+            model_name="rest.service.registration",
+            collection=collection,
+            request=request,
+            controller=self,
+        )
+        provider = work.component(usage=self._component_context_provider)
+        return provider._get_component_context()
 
     def make_response(self, data):
         if isinstance(data, Response):
--- a/base_rest/controllers/main.py
+++ b/base_rest/controllers/main.py
@@ -52,31 +52,79 @@ class RestControllerType(ControllerType):
                     "controller_class": cls,
                 }
             )
+            _logger.info(
+                "Added rest controller %s for module %s",
+                _rest_controllers_per_module[cls._module][-1],
+                cls._module,
+            )
 
 
 class RestController(Controller, metaclass=RestControllerType):
     """Generic REST Controller
 
-    This controller provides generic routes conform to commen REST usages.
-    You must inherit of this controller into your code to register your REST
-    routes. At the same time you must fill 2 required informations:
+    This controller is the base controller used by as base controller for all the REST
+    controller generated from the service components.
+
+    You must inherit of this controller into your code to register the root path
+    used to serve all the services defined for the given collection name.
+    This registration requires 2 parameters:
 
     _root_path:
     _collection_name:
 
+    Only one controller by _collection_name, _root_path should exists into an
+    odoo database. If more than one controller exists, a warning is issued into
+    the log at startup and the concrete controller used as base class
+    for the services registered into the collection name and served at the
+    root path is not predictable.
+
+    Module A:
+        class ControllerA(RestController):
+            _root_path='/my_path/'
+            _collection_name='my_services_collection'
+
+    Module B depends A:                               A
+        class ControllerB(ControllerA):             /  \
+            pass                                   B    C
+                                                  /
+    Module C depends A:                          D
+        class ControllerC(ControllerA):
+            pass
+
+    Module D depends B:
+        class ControllerB(ControllerB):
+            pass
+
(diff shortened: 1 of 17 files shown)


base_rest/controllers/main.py
```diff
@@ -44,48 +44,81 @@ def __init__(cls, name, bases, attrs):  # noqa: B902
...
+    controller generated from the service components.
+
+    You must inherit of this controller into your code to register the root path
+    used to serve all the services defined for the given the collection name.
+    your REST. This registration requires 2 parameters:
 
     _root_path:
     _collection_name:
 
+    Only one controller by _collection_name, _root_patch should exists into an
```
simahawk:
```suggestion
    Only one controller by _collection_name, _root_path should exists into an
```

base_rest/controllers/main.py
```diff
@@ -44,48 +44,81 @@ def __init__(cls, name, bases, attrs):  # noqa: B902
...
+    odoo database. If more than one controller exists, a warning is issued into
+    the log at startup and the concrete controller used as base class
+    for the services registered into the collection name and served at the
+    root path is not predictable.
```
simahawk, line 79: this is not clear, can you rephrase?

base_rest/controllers/main.py
```diff
@@ -44,48 +44,81 @@ def __init__(cls, name, bases, attrs):  # noqa: B902
...
+
+    The following properties can be specified to define comman properties to
+    applies to the generetated REST routes.
```
simahawk:
```suggestion
    The following properties can be specified to define common properties to
    apply to generated REST routes.
```

base_rest/models/rest_service_registration.py
```diff
@@ -107,15 +114,30 @@ def _update_auth_method_controller(self, controller_class):
...
+    def _apply_default_cors_if_not_set(self, controller_class, routing):
+        default_attr_name = "_default_cors"
+        if hasattr(controller_class, default_attr_name) and "cors" not in routing:
+            cors = getattr(controller_class, default_attr_name)
+            routing["cors"] = cors
+            if cors:
+                # add http method 'OPTIONS' required by cors if the route is
+                # restricted to specific method
+                if routing.get("methods"):
+                    routing["methods"].append("OPTIONS")
```
simahawk:
```suggestion
            if cors and "OPTIONS" not in routing.get("methods", []):
                # add http method 'OPTIONS' required by cors if the route is
                # restricted to specific method
                routing["methods"].append("OPTIONS")
```

simahawk (approved):
just code review, LG

base_rest/models/rest_service_registration.py
lmignon: `methods`is not required. If not specified, all the methods are allowed for the given route...
```if cors and "OPTIONS" not in routing.get("methods", ["OPTIONS"]):```

File: base_rest/controllers/main.py (hunk 2 above)
Comment (lmignon) [reply base_rest/controllers/main.py:79]: Is-it clearer now? (btw thank you for the review)

**Comment by simahawk** [comment]
can you squash?

**Comment by lmignon** [comment]
@simahawk squash done..

**Comment by simahawk** [comment]
good to merge?

**Comment by lmignon** [comment]
IMO yes @sbidoul, is it ok for you?

File: base_rest/models/rest_service_registration.py [hunk 5]
```diff
@@ -107,15 +109,29 @@ def _update_auth_method_controller(self, controller_class):
...
+        default_attr_name = "_default_" + attr_name
+        if hasattr(controller_class, default_attr_name) and attr_name not in routing:
+            routing[attr_name] = getattr(controller_class, default_attr_name)
+
+    def _apply_default_cors_if_not_set(self, controller_class, routing):
+        default_attr_name = "_default_cors"
+        if hasattr(controller_class, default_attr_name) and "cors" not in routing:
+            cors = getattr(controller_class, default_attr_name)
+            routing["cors"] = cors
+            if cors and "OPTIONS" not in routing.get("methods", ["OPTIONS"]):
```
sbidoul, line 136: why `routing.get("methods", ["OPTIONS"])` and not `routing.get("methods", [])` ?

lmignon: because if `methods` is not specified, all methods are allowed... https://github.com/odoo/odoo/blob/13.0/odoo/http.py#L441

base_rest/restapi.py
```diff
@@ -50,18 +50,16 @@ def decorator(f):
                 paths = [paths]
             if not isinstance(http_methods, list):
                 http_methods = [http_methods]
-            if cors and "OPTIONS" not in http_methods:
+            if kw.get("cors", False) and "OPTIONS" not in http_methods:
```
sbidoul: `kw.get("cors")`

base_rest/restapi.py
```diff
@@ -5,15 +5,13 @@
...
+from odoo import _, http
 from odoo.exceptions import UserError
 
 from .tools import cerberus_to_json
 
 
-def method(
-    routes, input_param=None, output_param=None, auth=None, cors=None, csrf=False
-):
+def method(routes, input_param=None, output_param=None, **kw):
```
sbidoul, line 14: explicit auth, cors, csrf made them easily discoverable.

lmignon: It's on purpose... It's the only way I found to know if these parameters have been explicitly provided or not. This information is required to know if the default parameters provided by the controller must be applied or not on a route.

sbidoul (approved):

base_rest/restapi.py
sbidoul: ah, ok, thanks

PR #184, opened 2021-07-30
[IMP] base_rest: Allow to pass the url parameters as kwargs

albariera:
This PR forward port of #148

### DIFF
--- a/base_rest/components/service.py
+++ b/base_rest/components/service.py
@@ -149,6 +149,16 @@ class BaseRestService(AbstractComponent):
                        to the method as keyword args.
         :return:
         """
+        if args:
+            _logger.debug(
+                "The use of *args will not be supported in 15.0. "
+                "All rest methods should be adapted to support "
+                "keyword arguments for path paramters.\n"
+                "Service: %s\n"
+                "Method: %s\n",
+                self._name,
+                method_name,
+            )
         method = getattr(self, method_name, object())
         params = params or {}
         secure_params = self._prepare_input_params(method, params)
--- a/base_rest/components/service.py
+++ b/base_rest/components/service.py
@@ -134,7 +134,7 @@ class BaseRestService(AbstractComponent):
             return result
         return output_param.to_response(self, result)
 
-    def dispatch(self, method_name, *args, params=None):
+    def dispatch(self, method_name, *args, params=None, **kwargs):
         """
         This method dispatch the call to the final method.
         Before the call parameters are processed by the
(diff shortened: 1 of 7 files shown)


etobella (approved):

lmignon:
@albariera rest_log overrides the `service._dispatch`. This addon should also be modified into order to be able to merge this PR.
If would also be fine to add a log.debug when args is used to say that `*args`will no more be supported in 15.0 All the rest methods must be adapted to support keyword arguments for path paramters.

albariera:
Done @lmignon! Thank you for your comment :blush:

base_rest/components/service.py
```diff
@@ -149,15 +149,23 @@ def dispatch(self, method_name, *args, params=None):
                        to the method as keyword args.
         :return:
         """
+        if args:
+            _logger.debug(
+                "The use of *args will not be supported in 15.0. "
+                "All rest methods should be adapted to support "
+                "keyword arguments for path paramters."
+            )
```
lmignon:
```suggestion
                "keyword arguments for path paramters.\n"
                "Service: %s\n"
                "Method: %s\n",
                self._name,
                method_name
            )
```

base_rest/restapi.py
```diff
@@ -38,6 +44,8 @@ def method(
     :param cors: The Access-Control-Allow-Origin cors directive value.
     :param bool csrf: Whether CSRF protection should be enabled for the route.
                       Defaults to ``False``
+    :param bool endpoint_params_to_kwargs: Pass the query parameters as keyword
+                arguments, following the same logic than Werkzeug
```
lmignon, line 48:
```suggestion
                arguments, following the same logic than Werkzeug
                Pay attention that this behavior will be the default one into 
                Odoo 15.
          
```

lmignon (changes requested):
Thank you for your patience and your hard work. Can you add a test  into base_rest for this new functionality.

base_rest/components/service.py
albariera: Done!

base_rest/restapi.py
albariera: Done too :)

albariera:
I have tested using the same logic as the tests already present. Is this what you expected? @lmignon

base_rest/tests/test_controller_builder.py
```diff
@@ -302,6 +310,21 @@ def _get_partner_schema(self):
...
+                "methods": ["POST"],
+                "auth": "user",
+                "cors": None,
+                "csrf": False,
+                "routes": [
+                    "/test_controller/partner/<int:invoice_id>"
+                    "/<string:partner_id>/change_partner"
+                ],
+            },
+        )
```
lmignon, line 326: I think it's possible to test that controller method dispatch correctly the request to the service... 

```python
 from odoo.addons.website.tools import MockRequest
 with MockRequest(self.env):
            res = controller().post_change_partner(invoice_id=...., partner_id=, ...)
            self.assert...

```

PR #188, opened 2021-08-25
[12.0][BK][13.0][FIX] datamodel: Error when NestedModel inside other field (Example: field.List())

lmignon:


lmignon:
see #159

lmignon:
merging since this pr is already approved in 13.0 and tests are green

PR #190, opened 2021-09-02
[14.0][IMP] base_rest_datamodel: allow to receive partial datamodels as params

fdegrave:
Forward port of PR https://github.com/OCA/rest-framework/pull/170

In particular, this feature allows one to use datamodels to make partial updates through `PUT` requests. In this case, the endpoint declaration would look like this:

```python
    @restapi.method(
        [(["/"], "PUT")],
        input_param=Datamodel("partner.info", partial=True),
        output_param=Datamodel("partner.info"),
    )
```

If `partial` is not set (or without this feature), the params sent by the client in the request must contain values for all the fields declared as `required` in the datamodel. Note that it uses an existing functionality of `marshmallow_objects`, as a call to `dump` of a `Datamodel` declared as `partial` would contain only the subset of fields provided in the params.

### DIFF
--- a/base_rest_datamodel/restapi.py
+++ b/base_rest_datamodel/restapi.py
@@ -12,21 +12,29 @@ from odoo.addons.base_rest import restapi
 
 
 class Datamodel(restapi.RestMethodParam):
-    def __init__(self, name, is_list=False):
+    def __init__(self, name, is_list=False, partial=None):
         """
 
-        :param name: The datamdel name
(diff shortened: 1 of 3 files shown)


lmignon (approved):
Nice feature. Thank you @fdegrave. LGTM (Code review)

lmignon:
@fdegrave Can you fix the conflicts so we can proceed to the merge?

PR #195, opened 2021-09-07
FIX: double specification on requirements

opa-cormaza:
FIX: double specification on requirements

PR #199, opened 2021-10-03
[14.0][FIX] datamodel: replace deprecated use of "missing" argument

davejames:
Deprecated use of "missing" argument leads to a warning in the Odoo logs when tests are run. This PR replaces "missing" with "load_default" as introduced in https://github.com/marshmallow-code/marshmallow/pull/1742

### DIFF
--- a/datamodel/tests/test_build_datamodel.py
+++ b/datamodel/tests/test_build_datamodel.py
@@ -71,15 +71,15 @@ class TestBuildDatamodel(DatamodelRegistryCase):
 
         class Datamodel1(Datamodel):
             _name = "datamodel1"
-            field_str1 = fields.String(missing="field_str1")
+            field_str1 = fields.String(load_default="field_str1")
 
         class Datamodel2(Datamodel):
             _inherit = "datamodel1"
-            field_str2 = fields.String(missing="field_str2")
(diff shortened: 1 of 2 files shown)


davejames:
pre-commit is failing because apispec is included in the external_dependencies in multiple places both with and without a version

image

I think it's outside the scope of this PR and I'm not sure what the best approach to deal with it is. I think the requirements.txt file is fine as it is

simahawk:
> pre-commit is failing because apispec is included in the external_dependencies in multiple places both with and without a version
> 
> image
> 
> I think it's outside the scope of this PR and I'm not sure what the best approach to deal with it is. I think the requirements.txt file is fine as it is

this is caused by https://github.com/OCA/rest-framework/pull/195  which should be rolled back because of this explanation https://github.com/acsone/setuptools-odoo/issues/59 

Feel free to revert it here.

simahawk (dismissed):
seems legit

davejames:
thanks @simahawk I've gone ahead and committed the auto-generated requirements.txt

simahawk:
@davejames no big deal in this case but next time use `git revert $hash` :wink:

simahawk (approved):

lmignon (approved):

PR #202, opened 2021-10-04
[14.0][IMP] base_rest: New auth method 'public_or_default'

lmignon:
A special auth method named 'public_or_default' can be used. In such a case when the HTTP route will be generated, the auth method will be computed from the '_default_auth' property defined on the controller with 'public_or_' as prefix.

ping @sebastienbeau @sbidoul

### DIFF
--- a/base_rest/models/rest_service_registration.py
+++ b/base_rest/models/rest_service_registration.py
@@ -118,7 +118,7 @@ class RestServiceRegistration(models.AbstractModel):
             routing = getattr(method, ROUTING_DECORATOR_ATTR, None)
             if not routing:
                 continue
-            self._apply_default_if_not_set(controller_class, routing, "auth")
+            self._apply_default_auth_if_not_set(controller_class, routing)
             self._apply_default_if_not_set(controller_class, routing, "csrf")
             self._apply_default_if_not_set(controller_class, routing, "save_session")
             self._apply_default_cors_if_not_set(controller_class, routing)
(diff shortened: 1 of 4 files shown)


sebastienbeau (approved):
LGTM code review. Thanks for the work !

base_rest/models/rest_service_registration.py
```diff
@@ -128,6 +128,17 @@ def _apply_default_if_not_set(self, controller_class, routing, attr_name):
...
             routing[attr_name] = getattr(controller_class, default_attr_name)
 
+    def _apply_default_auth_if_not_set(self, controller_class, routing):
+        default_attr_name = "_default_auth"
+        default_auth = getattr(controller_class, default_attr_name, None)
+        if default_auth:
+            if "auth" in routing:
+                auth = routing["auth"]
+                if auth == "public_or_default":
+                    routing["auth"] = "public_or_" + default_auth
```
simahawk: and where is this auth method handler defined?
I'm not sure to understand the purpose here :thinking:

sbidoul: @simahawk public_or_jwt auth method over here (https://github.com/OCA/server-auth/pull/294) with an explanation of the use case.

simahawk: I see. Can you add something to the README and maybe point to that example or add a demo route?

sbidoul: Can we check if `"_auth_method_public_or_" + default_auth` exists on `ir.http` ? Because, for instance, it does not make sense for auth_api_key where the client must always pass an api key (but sometimes does not pass PARTNER_EMAIL - but that is not managed in auth_api_key).

sbidoul (changes requested):
As explained, check that the auth method actually exists

base_rest/models/rest_service_registration.py
lmignon: @sbidoul done

sbidoul (approved):

PR #204, opened 2021-10-05
[14.0][IMP] authenticated_partner_id in ServiceContextProvider

qgroulard:
ServiceContextProvider now provides the authenticated_partner_id.

Default is None but the method _get_authenticated_partner_id() can be redefined to pass the authenticated_partner_id based on the authentication mechanism of your choice.
As done in base_rest_auth_jwt.

### DIFF
--- a/base_rest/components/service_context_provider.py
+++ b/base_rest/components/service_context_provider.py
@@ -9,9 +9,17 @@ class BaseRestServiceContextProvider(Component):
     _name = "base.rest.service.context.provider"
     _usage = "component_context_provider"
 
-    @property
-    def request(self):
-        return self.work.request
+    def __init__(self, work_context):
+        super().__init__(work_context)
+        self.request = work_context.request
+        # pylint: disable=assignment-from-none
+        self.authenticated_partner_id = self._get_authenticated_partner_id()
+
+    def _get_authenticated_partner_id(self):
+        return None
 
     def _get_component_context(self):
-        return {"request": self.request}
+        return {
+            "request": self.request,
+            "authenticated_partner_id": self.authenticated_partner_id,
+        }
--- a/base_rest/components/user_component_context_provider.py
+++ b/base_rest/components/user_component_context_provider.py
@@ -1,0 +1,12 @@
+# Copyright 2021 ACSONE SA/NV
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+
+from odoo.addons.component.core import AbstractComponent
+
+
+class AbstractUserAuthenticatedPartnerProvider(AbstractComponent):
+    _name = "abstract.user.authenticated.partner.provider"
+
+    def _get_authenticated_partner_id(self):
+        return self.env.user.partner_id.id
--- a/base_rest/models/ir_rule.py
+++ b/base_rest/models/ir_rule.py
@@ -1,0 +1,27 @@
+# Copyright 2021 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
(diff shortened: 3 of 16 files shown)


lmignon (approved):
Thank you for this very important improvement @qgroulard ! LGTM (Code review)

lmignon:
ping @sbidoul @PierrickBrun @simahawk

base_rest/components/user_component_context_provider.py
```diff
@@ -0,0 +1,12 @@
+# Copyright 2021 ACSONE SA/NV
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+
+from odoo.addons.component.core import AbstractComponent
+
+
+class AbstractUserComponentContextProvider(AbstractComponent):
+    _name = "abstract.user.component.context.provider"
```
simahawk:
```suggestion
    _name = "abstract.user.context.provider"
```
we know already that's a component, right? :)

base_rest/components/service_context_provider.py
```diff
@@ -13,5 +13,15 @@ class BaseRestServiceContextProvider(Component):
     def request(self):
         return self.work.request
 
+    @property
```
simahawk: IMO we should start overriding __init__ and stop using `@property`.
This is what I'm thinking of:
```
def __init__(self, work_context):
    super().__init__(work_context)
    self.request = work_context.request
    self.authenticated_partner = self._get_authenticated_partner()
```
and this is valid for all our basic component classes IMO.

simahawk (changes requested):

base_rest/components/user_component_context_provider.py
sbidoul: Except this component provide a work context for other components, so the keyword is ComponentContext ?

sbidoul: Or, we may want to align the naming with `abstract.auth.jwt.authenticated.partner.provider`. So `abstract.user.authenticated.partner.provider` ?

base_rest/components/service_context_provider.py
sbidoul: I don't know nearly enough about component, but this sounds reasonable to me at first glance.

lmignon:
@qgroulard Can you adapt your code accordint to the last comments so we can move forward with this PR?

base_rest/components/service_context_provider.py
```diff
@@ -9,9 +9,16 @@ class BaseRestServiceContextProvider(Component):
     _name = "base.rest.service.context.provider"
     _usage = "component_context_provider"
 
-    @property
-    def request(self):
-        return self.work.request
+    def __init__(self, work_context):
+        super().__init__(work_context)
+        self.request = work_context.request
+        self.authenticated_partner_id = self._get_authenticated_partner_id()
```
sbidoul, line 16: @qgroulard pre-commit/pylint is complaining. Seems you are hitting https://github.com/PyCQA/pylint/issues/2559.

`# pylint: disable=assignment-from-none` on this line may help silence it.

sbidoul (approved):

base_rest/models/ir_rule.py
```diff
@@ -0,0 +1,27 @@
...
+    """Add authenticated_partner_id in record rule evaluation context.
+
+    This come from the env context, which is populated by the base_rest service layer
+    context provider.
+    """
+
+    _inherit = "ir.rule"
+
+    @api.model
+    def _eval_context(self):
```
simahawk, line 17: nice trick!

simahawk (approved):

lmignon (approved):

PR #206, opened 2021-10-06
[14.0] [IMP] allow to use multipart/form-data format

PierrickBrun:
This allows to send files while also providing additional info (eg. attachment name,  res_id and res_model)

Here is an implementation: https://github.com/OCA/helpdesk/pull/270

### DIFF
--- a/base_rest/http.py
+++ b/base_rest/http.py
@@ -122,6 +122,9 @@ class HttpRestRequest(HttpRequest):
                 msg = "Invalid JSON data: %s" % str(e)
                 _logger.info("%s: %s", self.httprequest.path, msg)
                 raise BadRequest(msg)
+        elif self.httprequest.mimetype == "multipart/form-data":
+            # Do not reassign self.params
+            pass
         else:
             # We reparse the query_string in order to handle data structure
             # more information on https://github.com/aventurella/pyquerystring
--- a/base_rest/restapi.py
+++ b/base_rest/restapi.py
@@ -1,12 +1,14 @@
 # Copyright 2018 ACSONE SA/NV
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
+import abc
 import functools
+import json
 
 from cerberus import Validator
 
 from odoo import _, http
-from odoo.exceptions import UserError
+from odoo.exceptions import UserError, ValidationError
 
 from .tools import cerberus_to_json
 
@@ -109,7 +111,8 @@ def method(routes, input_param=None, output_param=None, **kw):
     return decorator
 
 
-class RestMethodParam(object):
+class RestMethodParam(abc.ABC):
+    @abc.abstractmethod
     def from_params(self, service, params):
         """
         This method is called to process the parameters received at the
@@ -121,7 +124,8 @@ class RestMethodParam(object):
         :return: Value into the format expected by the method
         """
 
-    def to_response(self, service, result):
+    @abc.abstractmethod
+    def to_response(self, service, result) -> http.Response:
         """
         This method is called to prepare the result of the call to the method
         in a format suitable by the controller (http.Response or JSON dict).
@@ -131,13 +135,20 @@ class RestMethodParam(object):
         :return: http.Response or JSON dict
         """
 
-    def to_openapi_query_parameters(self, service, spec):
+    @abc.abstractmethod
+    def to_openapi_query_parameters(self, service, spec) -> dict:
         return {}
 
-    def to_openapi_requestbody(self, service, spec):
+    @abc.abstractmethod
+    def to_openapi_requestbody(self, service, spec) -> dict:
         return {}
 
-    def to_openapi_responses(self, service, spec):
+    @abc.abstractmethod
+    def to_openapi_responses(self, service, spec) -> dict:
+        return {}
+
+    @abc.abstractmethod
+    def to_json_schema(self, service, spec, direction) -> dict:
         return {}
 
 
@@ -148,21 +159,27 @@ class BinaryData(RestMethodParam):
         self._mediatypes = mediatypes
         self._required = required
 
+    def to_json_schema(self, service, spec, direction):
+        return {
+            "type": "string",
+            "format": "binary",
+            "required": self._required,
+        }
+
     @property
     def _binary_content_schema(self):
         return {
-            mediatype: {
-                "schema": {
-                    "type": "string",
-                    "format": "binary",
-                    "required": self._required,
-                }
-            }
+            mediatype: {"schema": self.to_json_schema(None, None, None)}
             for mediatype in self._mediatypes
         }
 
-    def to_openapi_requestbody(self, services, spec):
+    def to_openapi_requestbody(self, service, spec):
         return {"content": self._binary_content_schema}
+
+    def to_openapi_query_parameters(self, service, spec):
+        raise NotImplementedError(
+            "BinaryData are not (?yet?) supported as query paramters"
+        )
 
     def to_openapi_responses(self, service, spec):
         return {"200": {"content": self._binary_content_schema}}
@@ -210,7 +227,7 @@ class CerberusValidator(RestMethodParam):
         raise SystemError(_("Invalid Response %s") % validator.errors)
 
     def to_openapi_query_parameters(self, service, spec):
-        json_schema = self.to_json_schema(service, "input")
+        json_schema = self.to_json_schema(service, spec, "input")
         parameters = []
         for prop, spec in list(json_schema["properties"].items()):
             params = {
@@ -239,11 +256,11 @@ class CerberusValidator(RestMethodParam):
         return parameters
 
     def to_openapi_requestbody(self, service, spec):
-        json_schema = self.to_json_schema(service, "input")
+        json_schema = self.to_json_schema(service, spec, "input")
         return {"content": {"application/json": {"schema": json_schema}}}
 
     def to_openapi_responses(self, service, spec):
-        json_schema = self.to_json_schema(service, "output")
+        json_schema = self.to_json_schema(service, spec, "output")
         return {"200": {"content": {"application/json": {"schema": json_schema}}}}
 
     def get_cerberus_validator(self, service, direction):
@@ -260,7 +277,7 @@ class CerberusValidator(RestMethodParam):
             return Validator(schema, purge_unknown=True)
         raise Exception(_("Unable to get cerberus schema from %s") % self._schema)
 
-    def to_json_schema(self, service, direction):
+    def to_json_schema(self, service, spec, direction):
         schema = self.get_cerberus_validator(service, direction).schema
         return cerberus_to_json(schema)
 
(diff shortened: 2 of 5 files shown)


PierrickBrun:
Hi @lmignon , can you review ?

lmignon:
Hi @PierrickBrun 
Thank you this usefull contrib. I'll try to review it in the next hours...

base_rest/http.py
```diff
@@ -122,6 +122,8 @@ def __init__(self, httprequest):
                 msg = "Invalid JSON data: %s" % str(e)
                 _logger.info("%s: %s", self.httprequest.path, msg)
                 raise BadRequest(msg)
+        elif self.httprequest.mimetype == "multipart/form-data":
+            pass
```
lmignon: These changes are useless

base_rest/restapi.py
```diff
@@ -128,7 +128,7 @@ def _binary_content_schema(self):
     def to_openapi_requestbody(self, service):
         return {"content": self._binary_content_schema}
 
-    def to_openapi_responses(self, service):
+    def to_openapi_responses(self, service, direction):
```
lmignon: Why do you add 'direction' as parameter?

base_rest/restapi.py
```diff
@@ -234,3 +234,66 @@ def get_cerberus_validator(self, service, direction):
     def to_json_schema(self, service, direction):
         schema = self.get_cerberus_validator(service, direction).schema
         return cerberus_to_json(schema)
+
+
+class FormDataPart(object):
+    def __init__(self, name=None, **kwargs):
```
lmignon: kwargs is useless

base_rest/restapi.py
```diff
@@ -234,3 +234,66 @@ def get_cerberus_validator(self, service, direction):
...
+
+
+class StringFormDataPart(FormDataPart):
+    def to_openapi_part_property(self, service, direction):
+        return {"type": "string"}
+
+
+class JsonFormDataPart(FormDataPart, CerberusValidator):
+    def __init__(self, name=None, schema=None, is_list=False):
+        self._name = name
```
lmignon: 'self._name' must be initialized by a call to super. IMO you should not inherit from `CerberusValidator`  but takes as parameter a validator providing the method `to_json_schema` de cette maniète, ce `JsonFormDataPart`pourra fonctionner avec tous les validateurs possibles supportant le json...

```python
class JsonFormDataPart(FormDataPart):
    def __init__(self, name, validator):
        super().__init__(name)
        self._validator = validator
        if not hasattr(validator, 'to_json_schema'):
            raise ...

        def to_openapi_part_property(self, service, direction):
            return self.validator.to_json_schema()
```
In fact I'm wondering if a generic FormDataPart with a `RestMethodParam` instance as parameter could be enough...

Note to Myself: `RestMethodParam` should be renamed ....

lmignon (changes requested):

base_rest/http.py
PierrickBrun: In fact it is useful because else it will execute this line: https://github.com/OCA/rest-framework/blob/14.0/base_rest/http.py#L128
Which will reassign `self.params` and thus erase `httprequest.files` and `httprequest.form` that were passed to params by https://github.com/odoo/odoo/blob/e5355a03aee110cb0686e4fbf3c81434e6a88bc2/odoo/http.py#L736

base_rest/restapi.py
PierrickBrun: Oops

base_rest/restapi.py
PierrickBrun: Thanks, way better like this !

Not sure about how to do it with a `RestMethodParam`

base_rest/restapi.py
```diff
@@ -234,3 +234,68 @@ def get_cerberus_validator(self, service, direction):
...
+        self._validator = validator
+        if not hasattr(validator, "to_json_schema"):
+            raise Exception("The validator should provide 'to_json_schema'")
+
+    def to_openapi_part_property(self, service, direction):
+        schema = self._validator.to_json_schema(service, direction)
+        return schema
+
+
+class MultipartFormData(RestMethodParam):
```
simahawk, line 355: docstring explaining usage?

simahawk (changes requested):
would be nice to have tests if possible

PierrickBrun:
> would be nice to have tests if possible

Done. The test uses Cerberus to validate the params and the helpdesk implementation gives an example on how to do it with Datamodel (https://github.com/akretion/helpdesk/blob/14.0-add-rest_api/helpdesk_mgmt_rest_api/services/attachment.py#L70 )

Also BinaryData is tested now too

PierrickBrun:
Hi @simahawk @lmignon 
Can we merge this ? I think every concern has been addressed

base_rest/restapi.py
```diff
@@ -333,3 +333,77 @@ def to_json_schema(self, service, direction):
...
+            "multipart/form-data": {
+                "schema": {
+                    "type": "object",
+                    "properties": self.to_openapi_properties(service, direction),
+                }
+            }
+        }
+
+    def from_params(self, service, params):
+        return params
```
lmignon, line 408: @PierrickBrun When do you validate the your params?

base_rest/restapi.py
```diff
@@ -333,3 +333,77 @@ def to_json_schema(self, service, direction):
...
+
+    def to_openapi_requestbody(self, service):
+        return {"content": self.to_multipart_content_schema(service, "input")}
+
+    def to_openapi_responses(self, service):
+        return {"200": {"content": self.to_multipart_content_schema(service, "output")}}
+
+    def to_response(self, service, result):
+        return result
```
lmignon: What do you expect as result? How do you validate-it?

base_rest/restapi.py
PierrickBrun: Fixed it

base_rest/restapi.py
PierrickBrun: I removed this possibility. i don't see a use case where we would want to return a multipart-formdata from the API so I think this would be better to not maintain it.

PierrickBrun:
Hi @lmignon ,

I've done the refactor to use RestMethodParam directly. It works very well ! There is a separate commit for it to ease the review

base_rest/restapi.py
```diff
@@ -333,3 +335,71 @@ def to_json_schema(self, service, direction):
...
+            raise ValidationError(_("You must provide a dict of RestMethodParam"))
+        self._parts = parts
+
+    def to_openapi_properties(self, service, direction):
+        properties = {}
+        for key, part in self._parts.items():
+            if isinstance(part, BinaryData):
+                properties[key] = part.to_json_schema()
+            else:
+                properties[key] = part.to_json_schema(service, direction)
```
lmignon: @PierrickBrun If we rely on a spécific method here with `to_json_schema`, The signature must be the same for all the params and must be declared into `RestMethodParam`. Moreover, we should enforce the method signature defined into `RestMethodParam' by taking advantage of the `abstractmethod` decorator available on abc.ABC class.

```python

import abc

...

class RestMethodParam(abc.ABC):
     @abc.abstractmehod
    def from_params(self, service, params) -> Any:
        """
        This method is called to process the parameters received at the
        controller. This method should validate and sanitize these paramaters.
        It could also be used to transform these parameters into the format
        expected by the called method
        :param service:
        :param request: `HttpRequest.params`
        :return: Value into the format expected by the method
        """

     @abc.abstractmehod
    def to_response(self, service, result) -> http.Response:
        """
        This method is called to prepare the result of the call to the method
        in a format suitable by the controller (http.Response or JSON dict).
        It's responsible for validating and sanitizing the result.
        :param service:
        :param obj:
        :return: http.Response or JSON dict
        """
     @abc.abstractmehod
    def to_openapi_query_parameters(self, service) > dict:
        ...

     @abc.abstractmehod
    def to_openapi_requestbody(self, service) -> dict:
        ...

     @abc.abstractmehod
    def to_openapi_responses(self, service) -> dict:
        ...

     @abc.abstractmehod
    def to_json_schema(service, direction) -> dict:
        ...
```

lmignon (changes requested):
@PierrickBrun You're on the right way... :smirk:

base_rest/restapi.py
PierrickBrun: @lmignon done

base_rest/restapi.py
```diff
@@ -333,3 +350,73 @@ def to_json_schema(self, service, direction):
...
+            res["multipart/form-data"]["schema"]["encoding"] = encoding
+        return res
+
+    def from_params(self, service, params):
+        for key, part in self._parts.items():
+            param = None
+            if isinstance(part, BinaryData):
+                param = part.from_params(service, params[key])
+            else:
+                # If the part is not Binary, it should be JSON
```
lmignon, line 397: A little bit too restrictive but not blocking at this stage...

lmignon (approved):
Thank you for all this work @PierrickBrun. LGTM (code review only)

hparfr (approved):

PierrickBrun:
Hi @simahawk Can you update your review ?

PierrickBrun:
@lmignon I rebased and fixed the pre-commit issue

base_rest/restapi.py
```diff
@@ -131,13 +135,20 @@ def to_response(self, service, result):
...
+    def to_openapi_requestbody(self, service) -> dict:
         return {}
 
-    def to_openapi_responses(self, service):
+    @abc.abstractmethod
+    def to_openapi_responses(self, service) -> dict:
+        return {}
+
+    @abc.abstractmethod
+    def to_json_schema(self, service, direction) -> dict:
```
PierrickBrun: @lmignon I have a problem with this line.

The function's signature is `to_json_schema(self, service, direction)` in cerberus and datamodel, but in pydantic it is `to_json_schema(self, spec)`.

Since I don't fully understand the base_rest_pydantic module's architecture I don't know what the correct answer to this problem should be

lmignon: I'll check...

lmignon: @PierrickBrun see #268

lmignon: @PierrickBrun We should unify  the signature method of `to_json_schema`to take the 3 paramters ] service, spec and direction as optional by default.

lmignon:
@PierrickBrun Can you do a rebase? Once it's done I'll make a pr on yours to fix the issue.

PierrickBrun:
> @PierrickBrun Can you do a rebase? Once it's done I'll make a pr on yours to fix the issue.

@lmignon done

lmignon:
@PierrickBrun done here https://github.com/akretion/rest-framework/pull/1

lmignon:
Thank you @PierrickBrun! Can you squash the commit (even mine) into 1 commit? Once it's done, I'll merge your changes and forward port to 15.0.

PierrickBrun:
> Thank you @PierrickBrun! Can you squash the commit (even mine) into 1 commit? Once it's done, I'll merge your changes and forward port to 15.0.

It's done, thanks for the patch

PierrickBrun:
> /ocabor merge minor

Sorry @lmignon but /ocabot has been mispelled, can you retype it ? :)

PR #207, opened 2021-10-07
[14.0][IMP] base_rest: Add new CerberusListValidator used to support array …

lmignon:
…of json validated by cerberus as request or response body
image

### DIFF
--- a/base_rest/restapi.py
+++ b/base_rest/restapi.py
@@ -188,18 +188,14 @@ class BinaryData(RestMethodParam):
 
 
 class CerberusValidator(RestMethodParam):
-    def __init__(self, schema=None, is_list=False):
+    def __init__(self, schema):
         """
 
         :param schema: can be dict as cerberus schema, an instance of
                        cerberus.Validator or a sting with the method name to
                        call on the service to get the schema or the validator
-        :param is_list: Not implemented. Should be set to True if params is a
-                        collection so that the object will be de/serialized
-                        from/to a list
         """
         self._schema = schema
-        self._is_list = is_list
 
     def from_params(self, service, params):
         validator = self.get_cerberus_validator(service, "input")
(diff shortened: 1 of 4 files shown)


lmignon:
ping @simahawk @sebastienbeau

lmignon:
ping @acsonefho @rousseldenis @kevinkhao

base_rest/restapi.py
```diff
@@ -188,18 +188,14 @@ def _to_http_response(self, result):
 
 
 class CerberusValidator(RestMethodParam):
-    def __init__(self, schema=None, is_list=False):
+    def __init__(self, schema, is_list=False):
```
simahawk:
```suggestion
    def __init__(self, schema, **kw):
```

simahawk (commented):
to be documented in README?

acsonefho (approved):

base_rest/restapi.py
lmignon: Why would you want to have '**kw' as an argument. IMO it's useless and not safe. Here we know what is needed for the class and these python classes are not extensible by other addons.

lmignon:
> to be documented in README?

Would be nice but IMO readme is not the right format to provide the documentation for the API. It's not allowed to declare subsections into readme parts and if we document this validator, we should also provides the documentation for the whole API which is outside the scope of this RP.

base_rest/restapi.py
simahawk: Yeah, not the best indeed but I thought you were willing to support other args, my bad.

Anyway `is_list` is useless too, why shall you keep it?
For backward compat? If so, then you should log is deprecated, no? Or at least add a comment explaining why is there.
If not, just remove it.

simahawk:
> > to be documented in README?
> 
> Would be nice but IMO readme is not the right format to provide the documentation for the API. It's not allowed to declare subsections into readme parts and if we document this validator, we should also provides the documentation for the whole API which is outside the scope of this RP.

AFAIK yes, you can have a sub section using `~~~`.
You already have an example in README regarding the normal validator, why should it be different?

lmignon:
@simahawk done

base_rest/restapi.py
lmignon: removed... (My bad, I had forgotten to delete it)

simahawk (approved):

simahawk:
remember to squash ;)

lmignon:
> remember to squash ;)

it's done now that your review is completed ;-)

PR #208, opened 2021-10-13
[14.0][IMP] base_rest: allow custom collection, ease reuse of some features

simahawk:


simahawk:
@lmignon with these changes and https://github.com/OCA/edi/pull/469 I'm able to register routes on the fly for a given app.
See it in action on the draft of [shopfloor.app model](https://github.com/OCA/wms/blob/25ffbef8e924fc36fd97a9b6d02a9d84e61a9ebd/shopfloor_base/models/shopfloor_app.py) (from https://github.com/OCA/wms/pull/316).

lmignon:
@simahawk it's interesting. I'm thinking about the overall architecture of base_rest and how to make it evolve in a way that it become easy to use, easy to explain and based on concepts related to the implementation of a service API. I see in all your code a lot of good technical tips to cover some new needs and provide a more powerful framework to implement REST APIs. Unfortunately, it reinforces the complexity of the current approach. I am not against merging such a change, however it is very likely that the approach in the future will be different. Especially with the introduction of dedicated models / inheritable class / specific properties to support the concepts of:
* Entry point: Model to register an entry point for an (a set of) application(s) -> should replace the RestController
* Application
* Service / Resource
* Method / Resource path
...

simahawk:
> @simahawk it's interesting. I'm thinking about the overall architecture of base_rest and how to make it evolve in a way that it become easy to use, easy to explain and based on concepts related to the implementation of a service API. I see in all your code a lot of good technical tips to cover some new needs and provide a more powerful framework to implement REST APIs. Unfortunately, it reinforces the complexity of the current approach. I am not against merging such a change, however it is very likely that the approach in the future will be different.

@lmignon well, I see your point. My take: the changes in this PR are - AFAIS - harmless and do not alter the current behavior. Nor it encourages to use base_rest in a different way -> no confusion or issue for future refactoring.

What I've done on shopfloor side simply makes sure we can reuse the current base_rest mechanism.
Which is all I need now to be able to provide a configurable backend for such an app.

>  Especially with the introduction of dedicated models / inheritable class / specific properties to support the concepts of:
>     * Entry point: Model to register an entry point for an (a set of) application(s) -> should replace the RestController
> 
>     * Application
>     * Service / Resource
>     * Method / Resource path

Well, the endpoint* modules go into that direction somehow. Still unclear how to plug/adapt/rework in the ctx of rest-framework. Eg: the shopfloor.app routing stuff could be moved to `rest.app` or `endpoint.app` model.
We should keep experimenting a discussing more, I think :)

At least, endpoint_route_handler allows to glue them all together in a - apparently - nice way. 
Of course it adds another bit in the loop but this is only if you want to have such an app, which for now it's just an example in the shopfloor ecosystem.

lmignon (approved):
LGTM (code review only)

Thank you @simahawk

TDu (approved):
LG, I don't get the purpose of every change, but if tests are passing.

PR #210, opened 2021-10-16
[14.0] [FIX] test: Remove merged PR OCA/connector/pull/393

hailangvn:
PR https://github.com/OCA/connector/pull/393 has been merged so the test requirement is not needed any more.


simahawk:
good catch :smile:

PR #214, opened 2021-11-03
base_rest: decouple rest route registration

simahawk:
REST requests are detected by checking the services registry.
This limits the possibility to register additional routes
as to be handled with the HttpRestRequest class.

Now the route registration is decoupled allowing 3rd parties
to register their own routes without loading them from a controller.

### DIFF
--- a/base_rest/models/rest_service_registration.py
+++ b/base_rest/models/rest_service_registration.py
@@ -27,6 +27,7 @@ from ..core import (
     RestServicesRegistry,
     _rest_controllers_per_module,
     _rest_services_databases,
+    _rest_services_routes,
 )
 from ..tools import _inspect_methods
 
@@ -222,6 +223,7 @@ class RestServiceRegistration(models.AbstractModel):
                     else None
                 )
                 services_registry[controller_def["root_path"]] = controller_def
+                self._register_rest_route(controller_def["root_path"])
                 if (
                     current_controller
                     and current_controller != controller_def["controller_class"]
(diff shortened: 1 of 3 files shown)


simahawk:
@lmignon  could you have a look at this pls?

base_rest/models/rest_service_registration.py
```diff
@@ -240,6 +242,13 @@ def _init_global_registry(self):
         _rest_services_databases[self.env.cr.dbname] = services_registry
         return services_registry
 
+    def _register_rest_route(self, route_path):
```
lmignon, line 245: @simahawk Be careful, it's not the indent of base_rest to provide a mechanism to allow third party addons to register their own dedicated routes. This kind of approach put a too heavy dependency between addons and technical aspect of this library. 'controller less' route is on the roadmap  next to:
* the entry point concept,
* python typing for the definition of restmethod (as in fastapi),
* refactoring of the openapi documentation generation to be generated at application or entry point level
* add support for odoo pydantic (https://github.com/OCA/rest-framework/pull/218/files -> could replace base_jsonify, cerberus, ....)
* a new abstraction layer on top of the actual components

I'm ok to merge this but this api could be removed into the next evolutions.

simahawk: > I'm ok to merge this but this api could be removed into the next evolutions. 

Fine, let's set a green light then :wink: 

I don't mind if it gets refactored.  Today I need this because of the patch on `get_request` to select the right request class. Do you think that patch is going to stay anyway?

lmignon: > I don't mind if it gets refactored. Today I need this because of the patch on `get_request` to select the right request class. Do you think that patch is going to stay anyway?

No motivation to remove-it... therefore it should stay.

lmignon (approved):

PR #218, opened 2021-11-07
[14.0][ADD] pydantic: allows pydantic use into Odoo

lmignon:
The marriage of the power of [Pydantic](https://pydantic-docs.helpmanual.io) and the modular approach of Odoo. And yes, it's possible to write modern python code within odoo :smirk: 

Not yet fully tested in a project....


pydantic/README.rst
```diff
@@ -0,0 +1,201 @@
...
+same and define the same fields and methods. They can be used indifferently into
+the code. All the logic will be provided by the aggregated class.
+
+.. code-block:: python
+
+    partner1 = PartnerInfo.construct()
+    partner2 = PartnerInfoWithCoordintate.construct()
+
+    assert partner1.__class__ == partner2.__class__
+    assert PartnerInfo.schema() == PartnerInfoWithCoordintate.schema()
```
sbidoul:
```suggestion
    assert PartnerInfo.schema() == PartnerInfoWithCoordinate.schema()
```

pydantic/README.rst
```diff
@@ -0,0 +1,201 @@
...
+
+.. note::
+
+    Since validation occurs on instance creation, it's important to avoid to
+    create an instance of a Pydantic class by usign the normal instance
+    constructor `partner = PartnerInfo(..)`. In such a case, if the class is
+    extended by an other addon and a required field is added, this code will
+    no more work. It's therefore a good practice to use the `cosntruct()` class
```
sbidoul:
```suggestion
    no more work. It's therefore a good practice to use the `construct()` method
```

pydantic/README.rst
```diff
@@ -0,0 +1,201 @@
...
+    method to create a pydantic instance.
```
sbidoul, line 93: I'd add an mention that adding required fields violates de Liskov substitution principle and should generally be avoided. This is certainly forbidden in requests data structures. When extending response data structures I can see why this is useful, to document new fields that are guaranteed to be present when extension addons are installed.

I wonder if we can collect example use cases for extending data structures.

pydantic/builder.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class PydanticClassesBuilder(omodels.AbstractModel):
+    """Build the component classes
+
+    And register them in a global registry.
+
+    Every time an Odoo registry is built, the know pydantic models are cleared and
+    rebuilt as well.  The pydantic classes are built by taking every models with
+    a ``_name`` and applying pydantic models with an ``_inherits`` upon them.
```
sbidoul: is _name still applicable to this version?

pydantic/context.py
```diff
@@ -0,0 +1,8 @@
+# Copyright 2021 ACSONE SA/NV
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl.html)
+
+# define context vars to hold the pydantic registry
+
+from contextvars import ContextVar
+
+pydantic_registry = ContextVar("pydantic_registry")
```
sbidoul:
```suggestion
odoo_pydantic_registry = ContextVar("pydantic_registry")
```

?

pydantic/utils.py
```diff
@@ -0,0 +1,72 @@
...
+
+    """
+
+    def get(self, key: Any, default: Any = None) -> Any:
+        res = getattr(self._obj, key, default)
+        if isinstance(self._obj, models.BaseModel) and key in self._obj._fields:
+            field = self._obj._fields[key]
+            if field.type in ["one2many", "many2many"]:
+                return list(res)
+        return res
```
sbidoul, line 68: Is there anything needed to convert False to None here ?

sbidoul (commented):
This is the beginning of a revolution :) Fantastic stuff.

pydantic/README.rst
lmignon: > I'd add an mention that adding required fields violates de Liskov substitution principle and should generally be avoided. This is certainly forbidden in requests data structures. When extending response data structures I can see why this is useful, to document new fields that are guaranteed to be present when extension addons are installed.

:+1: 

> I wonder if we can collect example use cases for extending data structures.

We could but it's a tedious work

pydantic/builder.py
lmignon: no..

pydantic/utils.py
lmignon: I'll add a test

pydantic/README.rst
```diff
@@ -0,0 +1,211 @@
...
+        name: str
+
+
+    class PartnerInfo(BaseModel):
+        street: str
+        street2: str = None
+        zip_code: str = None
+        city: str
+        phone: str = None
+        is_componay : bool = Field(None)
```
yvaucher, line 58:
```suggestion
        is_company : bool = Field(None)
```

pydantic/readme/USAGE.rst
```diff
@@ -0,0 +1,119 @@
...
+        name: str
+
+
+    class PartnerInfo(BaseModel):
+        street: str
+        street2: str = None
+        zip_code: str = None
+        city: str
+        phone: str = None
+        is_componay : bool = Field(None)
```
yvaucher:
```suggestion
        is_company : bool = Field(None)
```

gurneyalex:
Hello @lmignon  I'm testing pydantic on 15.0 using python 3.9. I get an error when I try to define a model with  a Literal attribute. 

```
class Order(BaseModel):
    warehouse: Literal["store", "olva", "spidy"]
    # lots of things commented out
```

Crashes the initialisation of the module:

```
  File "/odoo/src/odoo/service/server.py", line 1246, in preload_registries
  File "/odoo/src/odoo/modules/registry.py", line 87, in new
  File "/odoo/src/odoo/modules/loading.py", line 581, in load_modules
    model._register_hook()
  File "/odoo/external-src/rest-framework/pydantic/builder.py", line 50, in _register_hook
    self.build_registry(registry)
  File "/odoo/external-src/rest-framework/pydantic/builder.py", line 83, in build_registry
    registry.init_registry([m.name for m in graph])
  File "/odoo/external-src/rest-framework/pydantic/registry.py", line 221, in init_registry
    self.resolve_submodel_fields()
  File "/odoo/external-src/rest-framework/pydantic/registry.py", line 176, in resolve_submodel_fields
    cls._resolve_submodel_fields(registry=self)
  File "/odoo/external-src/rest-framework/pydantic/models.py", line 185, in _resolve_submodel_fields
    cls._resolve_submodel_field(field, registry)
  File "/odoo/external-src/rest-framework/pydantic/models.py", line 196, in _resolve_submodel_field
    if issubclass(field.type_, BaseModel):
  File "/odoo/external-src/rest-framework/pydantic/models.py", line 154, in __subclasscheck__
    return super().__subclasscheck__(subclass)
  File "/usr/lib/python3.9/abc.py", line 102, in __subclasscheck__
    return _abc_subclasscheck(cls, subclass)
TypeError: issubclass() arg 1 must be a class
``` 

Do you have an idea what I could be doing wrong here?

lmignon:
The problem seems to come from the subclass check when we try to check if
the type of an annotation is a subclass of BaseModel. I've to dig into the
code to find the right way to fix this UC.

lmi

lmignon:
@gurneyalex The issue is fixed into my last commit.

gurneyalex:
@lmignon another one that tricked me: I'm unable to call super() on a model which python-inherits from another model:

```
from odoo.addons.pydantic.models import BaseModel

class MyModel(BaseModel):
    value: int=2

    def method(self):
        return self.value

class DerivedModel(MyModel):  # no extends, I want two different models here
    value2: int=3

    def method(self):
        return super().method() + self.value2

```

when I call method() on DerivedModel, I get "TypeError: super(type, obj): obj must be an instance or subtype of type"

Am I missing something?

lmignon:
@gurneyalex [I added a test for your UC](https://github.com/OCA/rest-framework/pull/218/commits/9bf49018a507b61bd0e41cdb7508c4b3b226b5fe) but I'm not able to reproduce it.

gurneyalex:
> @gurneyalex [I added a test for your UC](https://github.com/OCA/rest-framework/pull/218/commits/9bf49018a507b61bd0e41cdb7508c4b3b226b5fe) but I'm not able to reproduce it.

weird... I ended up calling the parent method in the old fashioned way `ParentClass.method(self)` which worked (but speek of modern Python :rofl: )  I'll try to come up with a minimal example.

lmignon:
> > @gurneyalex [I added a test for your UC](https://github.com/OCA/rest-framework/pull/218/commits/9bf49018a507b61bd0e41cdb7508c4b3b226b5fe) but I'm not able to reproduce it.
> 

@gurneyalex I reproduced the problem this weekend. It does not appear to be on py36 but on py38. It requires a lot of changes to solve it but I have the solution. As soon as I have some time, I will make the changes.

> weird... I ended up calling the parent method in the old fashioned way `ParentClass.method(self)` which worked (but speek of modern Python rofl ) I'll try to come up with a minimal example.

This way of doing bypass the expected class hierarchy build by odoo when resolving the dependency graph...

lmignon:
Closing this PR since this functionality is now supported by the [`extendable-pydantic`](https://pypi.org/project/extendable_pydantic/) python package and the odoo's addon [`extendable`](https://github.com/acsone/odoo-addon-extendable/tree/14.0/extendable), You can take al look at https://github.com/OCA/event/pull/247 adding the new addon  `event_rest_api`

PR #219, opened 2021-11-19
[14.0] base_rest: api docs improvements

simahawk:
See atomic commits for details.

### DIFF
--- a/base_rest/apispec/rest_method_security_plugin.py
+++ b/base_rest/apispec/rest_method_security_plugin.py
@@ -9,9 +9,10 @@ class RestMethodSecurityPlugin(BasePlugin):
     APISpec plugin to generate path security from a services method
     """
 
-    def __init__(self, service):
+    def __init__(self, service, user_auths=("user",)):
         super(RestMethodSecurityPlugin, self).__init__()
         self._service = service
+        self._supported_user_auths = user_auths
 
     def init_spec(self, spec):
         super(RestMethodSecurityPlugin, self).init_spec(spec)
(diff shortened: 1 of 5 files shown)


JuMiSanAr (approved):

base_rest/apispec/rest_method_security_plugin.py
```diff
@@ -29,7 +29,7 @@ def operation_helper(self, path=None, operations=None, **kwargs):
         if not operations:
             return
         auth = routing.get("auth", self.spec._params.get("default_auth"))
-        if auth == "user":
+        if auth.startswith("user"):
```
lmignon: What's the motivation(s) behind this change? The strict equality is required otherwise we could match another kind of authentication that has nothing to do with the standard odoo "user" auth. If an addon introduce a new kind of authentication, you must create add your own plugin to cover this new auth as it's done in 
base_rest_auth_** plugins

lmignon (changes requested):
see comment

base_rest/apispec/rest_method_security_plugin.py
simahawk: the rationale is written in the commit msg:

> The backend might provide additional 'user' like auth policy.
> For instance, 'endpoint_route_handler' introduces 'user_endpoint'
> which raises an auth error instead of session expired error
> (which redirects to /web/login automatically).

I don't want to have a glue module just for this. Matching "user*" sounds fair as there's no non-user auth matching this naming in our ecosystem.

lmignon: Just because it doesn't exist doesn't mean we shouldn't anticipate it.

simahawk: better now?

lmignon (approved):

PR #221, opened 2021-11-22
[14.0] base_rest_pydantic: New addon to use pydantic model into base_rest

lmignon:
* uses the external python lib [extendable_pydantic](https://github.com/lmignon/extendable-pydantic) published on [pypi](https://pypi.org/project/extendable_pydantic/) 
* [x]  includes #246


base_rest_pydantic/restapi.py
```diff
@@ -0,0 +1,195 @@
...
+            for name, sch in definitions.items():
+                if name in spec.components.schemas:
+                    continue
+                spec.components.schema(name, sch)
+            schema = replace_ref_in_schema(schema, schema)
+            spec.components.schema(schema_name, schema)
+        return {"$ref": f"#/components/schemas/{schema_name}"}
+
+
+class PydanticModelList(PydanticModel):
```
lmignon, line 113: @sbidoul I wonder if I should delete this class.  Lists could be supported by the base class by parsing the type given as a parameter and the specific list processing arguments only used in the case of a list.
```python
from typing import List

...


    @restapi.method(
        [(["/", "/search"], "GET")],
        input_param=PydanticModel(PartnerSearchParam),
        output_param=PydanticModel(List[PartnerShortInfo]),
        auth="public",
    )
    def search(self, partner_search_param):
        """
        """
```

sbidoul: Do we actually need PydanticModel at all in the input/output param declaration ? Could it be

```
input_param=PartnerSearchParam, 
output_param=List[PartnerShortInfo],
```

and let the framework detect they derive from pydantic.BaseModel to do its magic ?

sbidoul: iow, PydanticModel and PydanticModelList could be internal to the base_rest framework with `restapi.method` wrapping input_param and output_param in PydanticModel when it detects they derive from pydantic.BaseModel.

lmignon: for sure...

sbidoul: I don't want to add more work on you. It can come as a future ergonomy improvement.

lmignon: I'll leave the code as is. It will be part of the next big evolution.

andreampiovesana (approved):
merge?

lmignon:
@andreampiovesana did you have the opportunity to test the module and the 'extendable-pydantic' lib?

lmignon:
@andreampiovesana It seems I've to fix the demo addon

base_rest_pydantic/i18n/base_rest_datamodel.pot
```diff
@@ -0,0 +1,26 @@
+# Translation of Odoo Server.
```
PierrickBrun, line 1: I don't know if this will break anything but this should be named `base_rest_pydantic.pot`

base_rest/restapi.py
```diff
@@ -161,10 +161,10 @@ def _binary_content_schema(self):
             for mediatype in self._mediatypes
         }
 
-    def to_openapi_requestbody(self, service):
+    def to_openapi_requestbody(self, services, spec):
```
PierrickBrun, line 164: I have the feeling that `services` is an error here and should be singular `service`

PR #225, opened 2021-11-24
[MIG][15.0] base_rest: Migration to 15.0

yankinmax:


yankinmax:
@simahawk @lmignon could you pls approve running workflow to proceed further?

oca_dependencies.txt
```diff
@@ -0,0 +1,2 @@
+# See https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#oca_dependencies-txt
```
simahawk: remove this

yankinmax: forgot about this, sorry

base_rest/__manifest__.py
```diff
@@ -6,18 +6,26 @@
...
-    "version": "14.0.4.3.1",
+    "version": "15.0.1.0.0",
     "development_status": "Beta",
     "license": "LGPL-3",
     "author": "ACSONE SA/NV, " "Odoo Community Association (OCA)",
     "maintainers": ["lmignon"],
     "website": "https://github.com/OCA/rest-framework",
     "depends": ["component"],
     "data": [
-        "views/assets_template.xml",
```
lmignon, line 17: Why do you remove this file? Without the assets, the openapi UI will no more work.

yankinmax: https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-15.0
```
Assets are now directly declared in the manifest. You should move .js and .scss links from templates to __manifest__.py file.
```

yankinmax: But the problem isn't with python code in checks. I can't understand what is wrong with README and index. Codecov check fails on that.

lmignon: OK my bad. Thank you for the link.

yankinmax: you're welcome

yankinmax:
@lmignon can you pls help, how to make it always running all check tests?

simahawk:
> @lmignon can you pls help, how to make it always running all check tests?

we cannot do anything about it AFAIK

yankinmax:
so I should always ask to approve workflow?

simahawk:
Reason https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

yankinmax:
> Reason https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks

thanks for the link

yankinmax:
@simahawk so, from what I see now, there is no codecov message if "Files changed". I need to improve tests.

lmignon (approved):
Thank you for the migration. LGTM (Code review only)

base_rest/http.py
```diff
@@ -2,7 +2,7 @@
 # Copyright 2017 Akretion (http://www.akretion.com).
 # @author Sébastien BEAU <sebastien.beau@akretion.com>
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
-
+# flake8: noqa
```
simahawk: reason for `noqa`?

simahawk (approved):

base_rest/http.py
yankinmax: pre-commit forces to "fix" here:
https://github.com/OCA/rest-framework/blob/feb7b9182bc5a4b5569e5f23411b259226276b48/base_rest/http.py#L121
Then if I add:
```
except ValueError as e:
      msg = "Invalid JSON data: %s" % str(e)
      _logger.info("%s: %s", self.httprequest.path, msg)
      raise BadRequest(msg) from e
```
It should be tested due to codecov, which seems not efficient for me. I tried actually, but couldn't succeed.
I tried to add `# flake8: disable=W1115` and it didn't work.
If this `noqa` isn't blocking I would leave it as it's.

simahawk: but here you here you are disabling it for the whole file. IMO it's wrong.
IMO is wrong to turn of the coverage check at all, unless is something really trivial.
Please roll it back and ignore the coverage color.

simahawk (changes requested):
rollback the noqa

simahawk (approved):

PR #234, opened 2021-12-10
[14.0][MIG] Migration to version 14.0: graphql_base

qgroulard:


graphql_base/__manifest__.py
```diff
@@ -5,14 +5,14 @@
...
         Base GraphQL/GraphiQL controller""",
-    "version": "13.0.1.0.0",
+    "version": "14.0.1.0.0",
     "license": "LGPL-3",
     "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
     "website": "https://github.com/OCA/rest-framework",
     "depends": ["base"],
     "data": ["views/graphiql.xml"],
-    "external_dependencies": {"python": ["graphene", "graphql-server-core"]},
+    "external_dependencies": {"python": ["graphene", "graphql-server", "promise"]},
```
sbidoul: As discussed, `promise` is used only by `graphene.test.Client` so this dependency is best declared in `test-requirements.txt`.

graphql_base/__manifest__.py
```diff
@@ -5,14 +5,14 @@
...
     "development_status": "Beta",
```
sbidoul: I think we can move this to `Production/Stable`

sbidoul (approved):
Code review and test.

lmignon (approved):

PR #236, opened 2021-12-14
[14.0][FIX] fix oca_dependencies.txt

marcelsavegnago:
https://github.com/OCA/rest-framework/blob/feb7b9182bc5a4b5569e5f23411b259226276b48/base_rest_demo/__manifest__.py#L14


lmignon:
@marcelsavegnago The oca dependencies should not be updated by hand. It should be generated by the pre-commit.

PR #238, opened 2021-12-15
[15.0][MIG] base_rest_auth_user_service: Migration to 15.0

yankinmax:


yankinmax:
@lmignon can we proceed here?

PR #240, opened 2021-12-29
[15.0][MIG] rest_log: Migration to 15.0

yankinmax:


yankinmax:
@simahawk can you pls take a look at this one?

simahawk:
@yankinmax pls port these changes too https://github.com/OCA/rest-framework/pull/243

rest_log/tests/test_db_logging.py
```diff
@@ -204,3 +228,102 @@ def test_log_entry_severity_mapping_param_bad_values(self):
...
+    @staticmethod
+    def _get_test_controller(class_or_instance, root_path=None):
+        return super()._get_test_controller(
+            class_or_instance, root_path="/test_log_exception_value/"
+        )
+
+    def test_log_exception_value(self):
+        self._test_exception(
+            "value", log_exceptions.RESTServiceDispatchException, "ValueError", "severe"
+        )
```
yankinmax, line 329: @simahawk these tests are failing and I don't have enough context about error log. Pls help to fix.
```
odoo/src/odoo/sql_db.py:89: in check
    return f(self, *args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <odoo.sql_db.Cursor object at 0x7f2a68d063a0>, query = 'ROLLBACK TO SAVEPOINT test_10', params = None, log_exceptions = None

    @check
    def execute(self, query, params=None, log_exceptions=None):
        if params and not isinstance(params, (tuple, list, dict)):
            # psycopg2's TypeError is not clear if you mess up the params
            raise ValueError("SQL query parameters should be a tuple, list or dict; got %r" % (params,))
    
        if self.sql_log:
            _logger.debug("query: %s", self._format(query, params))
        start = time.time()
        try:
            params = params or None
>           res = self._obj.execute(query, params)
E           psycopg2.errors.InvalidSavepointSpecification: savepoint "test_10" does not exist

odoo/src/odoo/sql_db.py:310: InvalidSavepointSpecification
```

simahawk: hmm the problem comes from the fact that now `TransactionCase` uses a savepoint,

These tests ensure that when an exception is catched while executing an endpoint, a rest.log entry is created.
In a SavepointCase the savepoint of the test case is rolled back immediately.

We could try using `SingleTransactionCase` which uses a single transaction w/out savepoint.

Option B: we could try using a savepoint [here](https://github.com/OCA/rest-framework/blob/14.0/rest_log/components/service.py#L42) as suggested by @lmignon in the past (see the comment right above).

Instead of using a try/except block we could use a savepoint and rollback to it. This way the original transaction is not invalidated. But: I'm not sure it will solve the problem and it implies a refactoring.

simahawk:
https://github.com/OCA/rest-framework/pull/261

PR #242, opened 2022-01-12
[ADD] 14.0 base rest abstract attachment

PierrickBrun:
based on #206


base_rest_abstract_attachment/__manifest__.py
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "REST API abstract attachment",
+    "summary": "Add an abstract component to manage attachments",
+    "version": "14.0.1.0.0",
+    "website": "https://github.com/OCA/rest-framework",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "license": "AGPL-3",
```
rvalyi: I think this kind of technical/framework module would be better licensed under LGPL-3 to get a broader adoption if you see no problem...

PierrickBrun: Indeed, no problem

sbidoul:
@qgroulard you may be interested in this one (for the images, maybe, or not)

base_rest_abstract_attachment/__manifest__.py
```diff
@@ -0,0 +1,16 @@
...
+    "version": "14.0.1.0.0",
+    "website": "https://github.com/OCA/rest-framework",
+    "author": "Akretion, Odoo Community Association (OCA)",
+    "license": "LGPL-3",
+    "application": False,
+    "installable": True,
+    "depends": [
+        "base_rest",
+        "base_jsonify",
+        "base_rest_datamodel",
```
lmignon: I would prefer to rely on pydantic in place of datamodel and base_jsonify.

base_rest_abstract_attachment/services/abstract_attachable.py
```diff
@@ -0,0 +1,158 @@
...
+
+    _name = "abstract.attachable.service"
+    _inherit = "base.rest.service"
+    _description = __doc__
+
+    def _json_parser_attachments(self):
+        res = [
+            ("attachment_ids:attachments", ["id", "name"]),
+        ]
+        return res
```
lmignon: this parser seems so be never used

base_rest_abstract_attachment/services/abstract_attachable.py
```diff
@@ -0,0 +1,158 @@
...
+        ),
+        output_param=restapi.Datamodel("ir.attachment.output"),
+    )
+    def create_attachment(self, object_id=None, file=None, params=None):
+        record = self._get(object_id)
+        self._check_attachment_access(record)
+        vals = self._prepare_attachment_params(record, file, params.dump())
+        attachment = self.env["ir.attachment"].create(vals)
+        return self.env.datamodels["ir.attachment.output"].load(
+            attachment.jsonify(["id", "name"])[0]
```
lmignon: ouchhh odoo.Model -> dict -> DataModel -> json... Do you really need all these steps.. You could achieve the same result by using the orm mode on pydantic models... see (https://github.com/OCA/rest-framework/pull/221/files#diff-3a30a8db1624833eda6491b3f0eb7a0569c3c713b1031062ee1b0df895f9ab98R48)

base_rest_abstract_attachment/services/abstract_attachable.py
```diff
@@ -0,0 +1,158 @@
...
+    _inherit = "base.rest.service"
+    _description = __doc__
+
+    def _json_parser_attachments(self):
+        res = [
+            ("attachment_ids:attachments", ["id", "name"]),
+        ]
+        return res
+
+    def _subvalidator_attachments(self):
```
lmignon: This validator seems to be never used

base_rest_abstract_attachment/services/abstract_attachable.py
PierrickBrun: I've migrated to pydantic now

base_rest_abstract_attachment/__manifest__.py
PierrickBrun: done

simahawk:
can't we name it only `base_rest_attachment`? I don't see the point of having "abstract" in the name :wink:

base_rest_abstract_attachment/readme/DESCRIPTION.rst
```diff
@@ -0,0 +1,3 @@
+This module is a technical module aiming to be used in other modules.
```
simahawk:
```suggestion
Technical module to be used as a base for other modules.
```

base_rest_abstract_attachment/services/abstract_attachable.py
```diff
@@ -0,0 +1,114 @@
...
+from odoo.addons.base_rest_pydantic.restapi import PydanticModel
+from odoo.addons.component.core import AbstractComponent
+
+from ..pydantic_models.attachment import AttachmentInfo, AttachmentRequest
+
+
+class AbstractAttachableService(AbstractComponent):
+    """Abstract service to allow to use attachments on a service."""
+
+    _name = "abstract.attachable.service"
```
simahawk:
```suggestion
class RestAttachmentServiceMixin(AbstractComponent):
    """Abstract service to allow to use attachments on a service."""

    _name = "rest.attachment.service.mixin"
```

base_rest_abstract_attachment/services/abstract_attachable.py
```diff
@@ -0,0 +1,114 @@
...
+    _inherit = "base.rest.service"
+    _description = __doc__
+
+    @restapi.method(
+        routes=[(["/<int:object_id>/attachments/<int:attachment_id>"], "GET")],
+        output_param=restapi.BinaryData(required=True),
+    )
+    def download_attachment(self, object_id=None, attachment_id=None):
+        record = self._get(object_id)
+        self._check_attachment_access(record)
```
simahawk: the check should be called by `_get` IMO

base_rest_abstract_attachment/services/abstract_attachable.py
```diff
@@ -0,0 +1,114 @@
...
+        attachment = self._get_attachment_for_record(record, attachment_id)
+        content = None
+        headers = [("X-Content-Type-Options", "nosniff")]
+        if attachment.type == "url":
+            r = requests.get(attachment.url)
```
simahawk:
```suggestion
        content, headers = self._get_attachment_response_data(attachment)
```
and move everything there

base_rest_abstract_attachment/tests/test_attachment.py
```diff
@@ -0,0 +1,60 @@
...
+        with open(pathlib.Path(__file__).resolve()) as fp:
+            attrs["file"] = FileStorage(fp)
+            res = self.service.dispatch("create_attachment", params=attrs)
+        return res
+
+
+class AttachmentCase(
+    AttachmentCommonCase, TransactionRestServiceRegistryCase, ExtendableMixin
+):
+    def setUp(self):
```
simahawk: you can use the Savepoint case and run this only at class setup

PierrickBrun:
@simahawk You're right I've renamed the module.

Thanks for the review, I've made the changes, the tests still do not pass, I will need to look into it again to make it work

PierrickBrun:
Also, I've used the same strategy as base_jsonifier to rename the module without breaking the installed databases so there is now and empty module inheriting the not empty one

simahawk:
@PierrickBrun shall we reopen this?

PR #246, opened 2022-02-04
[14.0][ADD] extendable: New addon to alllow usage of the Extendable python …

lmignon:
…package into odoo

ping @sbidoul @qgroulard


extendable/README.rst
```diff
@@ -0,0 +1,96 @@
...
+
+**Table of contents**
+
+.. contents::
+   :local:
+
+Usage
+=====
+
+By default, all extendable classes defined into your odoo addons are automatically
```
sbidoul, line 40:
```suggestion
By default, all [extendable](https://pypi.org/project/extendable/) classes defined into your odoo addons are automatically
```

sbidoul (approved):
Fantastic approach to create *pythonic* classes that are extendable in Odoo addons.

First use here: https://github.com/OCA/event/pull/247

marcelsavegnago (approved):

rvalyi:
is [extendable](https://github.com/lmignon/extendable) "Odoo done right" ;-) ?

lmignon:
What do you mean by `Odoo done right` ? The [extendable](https://pypi.org/project/extendable) python package lack of documentation to expose the principles, concepts and implementation of this package (will be done in the future). In short, it allows you to write normal/modern python code with the same extension mechanism as the odoo models. With extendable, you always use your class definition in your code (never an object from a registry). This class act as a proxy to the concrete class.  This concrete class is the result of the aggregation of all the specializations defined into the code and known at runtime. It's thus possible to write your python code quite conveniently and even more because this approach supports python static typing. This makes the code easier to write (auto completion) and maintain. It also becomes possible to use mypy to check your code.  We are in the early stages of this approach and we are starting to use it in one project in order to confront it with concrete cases. It has taken a long time to get to this point, but I am confident that the foundations are solid enough now to make this work public (even if we don't advertise it.). I also tried to make sure that this library offers sufficient facilities to be able to mix other existing types in order to make them extendable. [extendable-pydantic](https://pypi.org/project/extendable_pydantic/) is the first illustration. (I can even say that it is the will to be able to make pydantic extedable "like" odoo models that was the starting point of all this)

PR #249, opened 2022-02-11
[13.0][FIX] base_rest: correct external-dependency

MiquelRForgeFlow:
Use correct name, the one that is in requirements.

### DIFF
--- a/base_rest/__manifest__.py
+++ b/base_rest/__manifest__.py
@@ -20,7 +20,7 @@
     ],
     "demo": [],
     "external_dependencies": {
-        "python": ["cerberus", "pyquerystring", "accept_language", "apispec"]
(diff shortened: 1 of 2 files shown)


lmignon (approved):

lmignon (changes requested):
@MiquelRForgeFlow Thank you for the change. In the same time could you remove the corresponding line into the dependencies override of the setup.py file https://github.com/OCA/rest-framework/blob/13.0/setup/base_rest/setup.py#L8

MiquelRForgeFlow:
Done. Should I do the same for 12.0 and 11.0 branches?

sbidoul:
No this only works in 13 and later.

PR #147, opened 2021-04-21
[13.0] rest_log misc improvements

simahawk:
See atomic commits for details.

Most important change: you can now selectively enable logging by:

- `collection_name`  # enable for all endpoints of the collection
- `collection_name.usage`  # enable for specific endpoints
- `collection_name.usage.endpoint`  # enable for specific endpoints
- `collection_name*:state`  # enable only for specific state (success, failed)

PR #250, opened 2022-02-14
[14.0][FIX] rest_log: inject log entry URL only for dict response

davejames:
Fixes a regression introduced in https://github.com/OCA/rest-framework/pull/147/commits/cb4dfad3eab4803d1183e1b8d64c2cbcea6f8cda

Original fix: https://github.com/OCA/rest-framework/commit/c607a423a27a1d4a86a88952f4537599c7dff849

### DIFF
--- a/rest_log/components/service.py
+++ b/rest_log/components/service.py
@@ -68,7 +68,7 @@ class BaseRESTService(AbstractComponent):
         log_entry = self._log_call_in_db(
             self.env, request, method_name, *args, params=params, result=result
         )
-        if log_entry:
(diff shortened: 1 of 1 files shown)


simahawk:
@davejames thanks for your fix, I definitely screwed my fix during rebase/cherry-pick :sweat_smile: 
Could please cherry-pick my commit instead of replicating it?
Also, if you have time, you could add a test for it. Not mandatory of course :wink:

simahawk:
thanks for pointing out the issue, speeding this up in #251

PR #253, opened 2022-02-20
[13.0][MIG] base_rest_pydantic: backport

pierre-halleux:


lmignon:
@pierre-halleux It would be nice to fix this PR...

PR #258, opened 2022-04-05
fix args order mistake when the path has multi path arguments,

moebiusever:
eg.,rule with "/path/<string:in1>/<string:in2>" ,
path with /path/1/2,
will get in2=1, in1=2 with mistake sometimes
becuase the rule.arguments is type of set which was not ordered,
after fix, will always got in1=1, in2=2

### DIFF
--- a/base_rest/http.py
+++ b/base_rest/http.py
@@ -123,7 +123,6 @@ class HttpRestRequest(HttpRequest):
             except ValueError as e:
                 msg = "Invalid JSON data: %s" % str(e)
                 _logger.info("%s: %s", self.httprequest.path, msg)
-                raise BadRequest(msg)
         else:
             # We reparse the query_string in order to handle data structure
             # more information on https://github.com/aventurella/pyquerystring
--- a/base_rest/models/rest_service_registration.py
+++ b/base_rest/models/rest_service_registration.py
@@ -333,7 +333,7 @@ class RestApiServiceControllerGenerator(object):
                         method_name=method_name,
                         service_name=self._service_name,
                         service_method_name=name,
-                        args=", ".join(rule.arguments),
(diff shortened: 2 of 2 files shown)


etobella:
The error was solved with this https://github.com/OCA/rest-framework/pull/148 but it was never merged...

lmignon:
@etobella I prefer this far less intrusive approach proposed into this PR

etobella:
Well, I don't know if this really solves the issue, it might be. However, my PR fixed the approach, that it might be improved IMO. Logic was complicated in order to remove problems on existing installations, but it could be simplified in a newer version.

lmignon:
@etobella I prefer to not merge your pr since it adds a lot of complexity for an issue that can be fixed in 1 line change.

lmignon:
@moebiusever Welcome into the community. Thank you for the fix. Can you sign the Contributor License Agreement so we can proceed with the merge.

Regards

etobella:
@lmignon This is a patch, my solution is a fix and improvement. It is right for me if you prefer to patch it. However, the real error is still there...

My point is, the patch is still using arguments, we should use kwargs.

lmignon (approved):

moebiusever:
> @moebiusever Welcome into the community. Thank you for the fix. Can you sign the Contributor License Agreement so we can proceed with the merge.
> 
> Regards

@lmignon  Hello,  I‘ve signed it and sent to cla AT odoo-community.org. Sorry for my poor english.

lmignon:
@moebiusever your commit 2a81ec271f9037010c11203c7f327528d6a0ee46 has been merged in #264 The other one change the current behavior without motivation. I close this PR. Thank you for the fix

PR #259, opened 2022-04-06
[FIX] Fix to validate OpenAPI 3.0

arobertlobo5:
- http code 200 with description OK.
- JWT schema without name and without description


lmignon:
@arobertlobo5 Thank you for this PR. Can you explain why you remove the name and description fields from the JWT definition?

PR #262, opened 2022-06-01
base_rest: decorate service endpoints always

simahawk:
Decorate services' methods when the component is initialized.
This way, the state of the component is the same no matter where you get it from.

@lmignon what do you think?

Note that I left it in `rest.service.registration` but it can be moved in utils or to a specific component.

### DIFF
--- a/base_rest/components/service.py
+++ b/base_rest/components/service.py
@@ -56,6 +56,14 @@ class BaseRestService(AbstractComponent):
     _description = None  # description included into the openapi doc
     _is_rest_service_component = True  # marker to retrieve REST components
 
+    def __init__(self, work_context):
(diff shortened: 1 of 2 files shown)


lmignon:
@simahawk I'm not in favor of this change. The goal of the initial implementation is to decorate the service's methods when loading the registry. Therefore the code is called 1 time by service when a worker is initialized.  Moreover the decorators are applied at class level. With your change, the code will be called each time a service is initialized (on every request) and the decorator will be applied at the instance level. 
For Odoo16 I would like to remove the support for non decorated methods and therefore remove this code. (I'm also thinking of removing the dependency on "component" which would greatly simplify the stack)

PR #264, opened 2022-06-09
[FIX] 14.0 routing: method arguments ordered

PierrickBrun:
Before:

When using routing with several arguments: 
Arguments order can change on server reboot (werkzeug.routing.Rule argument parameter is a Set)

After:

arguments ordering is preserved

### DIFF
--- a/base_rest/models/rest_service_registration.py
+++ b/base_rest/models/rest_service_registration.py
@@ -411,7 +411,7 @@ class RestApiServiceControllerGenerator(object):
                         method_name=method_name,
                         service_name=self._service_name,
                         service_method_name=name,
-                        args=", ".join(rule.arguments),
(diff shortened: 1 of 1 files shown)


PierrickBrun:
I found out about https://github.com/OCA/rest-framework/pull/258 right after creating this PR.

Looks like it does exactly the same thing but with less complexity, I'll look into it

lmignon:
@PierrickBrun If #258 is OK, you could only cherry-pick https://github.com/OCA/rest-framework/pull/258/commits/2a81ec271f9037010c11203c7f327528d6a0ee46 in your PR and I'll merge it ASAP

PierrickBrun:
@lmignon Yes done

lmignon (approved):

PR #268, opened 2022-06-14
[14.0][FIX] base_rest_pydantic: Fix method signature for 'RestMethodParam.to_json_schema'

lmignon:


lmignon:
@PierrickBrun Is-it ok for you?

PR #269, opened 2022-06-15
[15.0] [IMP] allow to use multipart/form-data format

lmignon:
forward port of #206


lmignon:
fast tracking since the original PR has already been approved on 14.0

PR #270, opened 2022-06-15
[IMP] Deprecate old API

lmignon:
At startup, output deprecation message for methods not explicitely declared with the new api decorator

To help adapt the code, the decorator to add to the method not declared with the new api decorator is printed into the log file.
```bash
2022-06-15 15:02:07,093 535175 WARNING odoo-14-test odoo.addons.base_rest.models.rest_service_registration: "Deprecated REST API implicit method "session_expired" in service "exception.service"
The support for implicit rest method will be removed in 16.0
You should consider to explicitly declare your method as a rest method by adding the following decorator
@restapi.method(
    [('/session_expired', 'POST')],
    input_param=restapi.CerberusValidator(schema='_validator_session_expired'),
    output_param=restapi.CerberusValidator(schema='_validator_return_session_expired')
)
def session_expired(...):

```

### DIFF
--- a/base_rest/models/rest_service_registration.py
+++ b/base_rest/models/rest_service_registration.py
@@ -292,6 +292,41 @@ class RestApiMethodTransformer(object):
             routes=routes, input_param=input_param, output_param=output_param
         )(getattr(self._service.__class__, method_name))
         setattr(self._service.__class__, method_name, decorated_method)
+        # flake8: noqa: B950
+        deprecated_message = (
+            'Deprecated REST API implicit method "%s" '
+            'in service "%s". The support for'
+            " implicit rest method will be removed in 16.0. "
+            "For full details on how to adapt your code, please "
+            "enable DEBUG level"
+        )
+        _logger.warning(deprecated_message, method_name, self._service._name)
+        deprecated_message_details = """Add the following decorator to fix this issue:
+@restapi.method(
+    {params}
+)
+def {method_name}(...):
+        """
+        params = [str(routes)]
+        if input_param:
+            params.append(
+                "input_param=restapi.CerberusValidator(schema='{schema}')".format(
+                    schema=input_param._schema
+                )
+            )
+        if output_param:
+            params.append(
+                "output_param=restapi.CerberusValidator(schema='{schema}')".format(
+                    schema=output_param._schema
+                )
+            )
+        _logger.debug(
+            deprecated_message_details.format(
+                method_name=method_name,
+                service_name=self._service._name,
+                params=",\n    ".join(params),
+            )
+        )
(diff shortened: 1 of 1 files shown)


lmignon:
ping @simahawk @sbidoul

lmignon:
@sbidoul @simahawk The deprecation message is logged as warning. This makes runboat build fails. Do I change the level to info?

sbidoul:
That's not the reason for the build failure. I was connected to the template1 db to show something to Pierrick. Please retry.

simahawk (changes requested):
I wouldn't log the full info. Can we move details to `debug` level?

lmignon:
@simahawk I'm not sure that logging the details at debug level is a good idea. Personally, I only activate the debug level for some specific addons in some specific cases when we encounter unexpected behavior in production. Moreover, developers will never think about activating the debug level to get the details that will help them to adapt their code. If you are annoyed by the warnings, it's easy to fix your code with the details into the log message to get rid of them.

simahawk:
> @simahawk I'm not sure that logging the details at debug level is a good idea. Personally, I only activate the debug level for some specific addons in some specific cases when we encounter unexpected behavior in production. Moreover, developers will never think about activating the debug level to get the details that will help them to adapt their code. If you are annoyed by the warnings, it's easy to fix your code with the details into the log message to get rid of them.

I'm not saying you shouldn't see anything. I'm saying log the deprecation w/out the detail in INFO mode and mention that more details can be seen w/ DEBUG mode. Eg:

"""
Deprecated REST API implicit method "session_expired" in service "exception.service". The support for implicit rest method will be removed in 16.0. For full details please enable DEBUG level
"""
IMO logging the same thing hundreds time won't help devs more.

lmignon:
@simahawk And once you enable the debug log level the information you're looking for is lost in a tons of useless log entries....

lmignon:
done
```
2022-06-16 08:56:30,589 574334 WARNING odoo-14-test odoo.addons.base_rest.models.rest_service_registration: Deprecated REST API implicit method "get" in service "partner_image.service". The support for implicit rest method will be removed in 16.0. For full details on how to adapt your code, please enable DEBUG level 
2022-06-16 08:56:30,590 574334 DEBUG odoo-14-test odoo.addons.base_rest.models.rest_service_registration: Add the following decorator to fix this issue:
@restapi.method(
    [(['/<int:id>/get', '/<int:id>'], 'GET')],
    input_param=restapi.CerberusValidator(schema='_validator_get')
)
def get(...):
```

lmignon:
@simahawk Can you update your review PLZ? Thank you

Issue #154, opened 2021-05-08
Patched version Root.get_request was not called for the very first HTTP request


xu4wang:
There is an issue regarding the very first HTTP call after odoo restart. It is always handled by the original get_request handler, instead the one Rest Framework patched as below.

```
def get_request(self, httprequest):
    db = httprequest.session.db
    if db:
        # on the very first request processed by a worker,
        # registry is not loaded yet
        # so we enforce its loading here to make sure that
        # _rest_services_databases is not empty
        odoo.registry(db)
        service_registry = _rest_services_databases.get(db)
        if service_registry:
            for root_path in service_registry:
                if httprequest.path.startswith(root_path):
                    return HttpRestRequest(httprequest)
    return ori_get_request(self, httprequest)


Root.get_request = get_request
```

For the debug message, the Root.get_request was indeed patched(2021-05-08 17:51:45,967) after base_rest loaded.  However the next HTTP request was still handled by the original odoo handler(2021-05-08 17:51:46,191)

```
2021-05-08 17:51:45,904 42021 DEBUG gateway odoo.modules.loading: loading module datamodel (2/13)
2021-05-08 17:51:45,918 42021 DEBUG gateway odoo.modules.loading: loading module component (3/13)
2021-05-08 17:51:45,926 42021 DEBUG gateway odoo.modules.loading: loading module web (4/13)
2021-05-08 17:51:45,930 42021 DEBUG gateway odoo.modules.loading: loading module base_import (5/13)
2021-05-08 17:51:45,950 42021 DEBUG gateway odoo.modules.loading: loading module base_rest (6/13)
2021-05-08 17:51:45,967 42021 DEBUG gateway odoo.addons.base_rest.http: --------Patching Root.get_request
2021-05-08 17:51:45,967 42021 DEBUG gateway odoo.modules.loading: loading module odoo_referral (7/13)
2021-05-08 17:51:45,970 42021 DEBUG gateway odoo.modules.loading: loading module web_diagram (8/13)
2021-05-08 17:51:45,972 42021 DEBUG gateway odoo.modules.loading: loading module web_editor (9/13)
2021-05-08 17:51:45,990 42021 DEBUG gateway odoo.modules.loading: loading module web_kanban_gauge (10/13)
2021-05-08 17:51:45,991 42021 DEBUG gateway odoo.modules.loading: loading module web_tour (11/13)
[...]

PR #276, opened 2022-08-10
[FIX] base_rest: Patch from the first request

SirTakobi:
This is to help users solving issue https://github.com/OCA/rest-framework/issues/154.

I have reproduced this issue locally with the steps:

1. Install `base_rest_demo`
2. Open the REST api app, ensure this is the only window opened that is contacting Odoo
3. Choose the new api partner services and do a search -> this works fine
4. Reboot odoo, without closing the window
5. Choose the new api partner services and do a search -> this raises

    >  File "/home/simone/work/odoo12/rest-framework/base_rest/controllers/main.py", line 95, in make_response
    return request.make_json_response(data)
  File "/home/simone/work/odoo12/venv/lib/python3.6/site-packages/werkzeug/local.py", line 343, in __getattr__
    return getattr(self._get_current_object(), name)
AttributeError: 'HttpRequest' object has no attribute 'make_json_response' - - -
6. Choose the new api partner services and do a search -> this works fine again

The configuration patch has been suggested in:
>This patch work like a charm ...! 
>
>You need to load base_rest at odoo start. Because, this enforce the loading of the registry in the get_request().
>So we can use base_rest as server_wide_module()
>
>Here odoo conf example :
>```
>[options]
># ex: for Odoo 14.0
>server_wide_modules=base,web,base_rest
># ...
>
_Originally posted by @dinuth-perera in https://github.com/OCA/rest-framework/issues/154#issuecomment-1053658207_

### DIFF
--- a/base_rest/controllers/main.py
+++ b/base_rest/controllers/main.py
@@ -6,7 +6,8 @@ from contextlib import contextmanager
 
 from werkzeug.exceptions import BadRequest
 
-from odoo.http import Controller, ControllerType, Response, request
+from odoo.http import Controller, ControllerType, Response, request, \
+    controllers_per_module
 
 from odoo.addons.component.core import WorkContext, _get_addon_name
 
@@ -36,6 +37,13 @@ class RestControllerType(ControllerType):
             # our RestConrtroller must be a direct child of Controller
             bases += (Controller,)
         super(RestControllerType, cls).__init__(name, bases, attrs)
+        # The generic controller should not be registered as a controller
+        # even if it inherits from Controller
+        base_rest_controllers = controllers_per_module['base_rest']
+        name_class = ("%s.%s" % (cls.__module__, 'RestController'), cls)
+        if name_class in base_rest_controllers:
+            base_rest_controllers.remove(name_class)
+
         if "RestController" not in globals() or not any(
             issubclass(b, RestController) for b in bases
         ):
--- a/base_rest/readme/CONFIGURE.rst
+++ b/base_rest/readme/CONFIGURE.rst
@@ -1,3 +1,9 @@
+Load this module at Odoo startup including in your server config file:
+
+.. code-block:: cfg
+
+    server_wide_modules=base,web,base_rest
+
 If an error occurs when calling a method of a service (ie missing parameter,
 ..) the system returns only a general description of the problem without
 details. This is done on purpose to ensure maximum opacity on implementation


simahawk:
Should we load the patch w/ post load hook? This should grant the patch to be loaded when the registry gets initialized AFAIK

SirTakobi:
Thanks for having a look!
> Should we load the patch w/ post load hook? This should grant the patch to be loaded when the registry gets initialized AFAIK

I don't know how to do that, do you have any suggestion?

I had a look at the code and I don't know if that would work anyway because the problem is that the Class of the first request is decided before loading any module in https://github.com/odoo/odoo/blob/67461b4ca1a9c2eb8ea1c0cc4bc39d91cf0d2823/odoo/http.py#L1456
Loading the modules starts a few lines below in https://github.com/odoo/odoo/blob/67461b4ca1a9c2eb8ea1c0cc4bc39d91cf0d2823/odoo/http.py#L1471
and after that, any request is correctly patched by
https://github.com/OCA/rest-framework/blob/c417c7eaf7aa7ee4b9c880d326936322365060b3/base_rest/http.py#L222

simahawk:
> Thanks for having a look!
> 
> > Should we load the patch w/ post load hook? This should grant the patch to be loaded when the registry gets initialized AFAIK
> 
> I don't know how to do that, do you have any suggestion?

Just search for `post_load` in OCA modules, you'll find some examples.

> I had a look at the code and I don't know if that would work anyway because the problem is that the Class of the first request is decided before loading any module in https://github.com/odoo/odoo/blob/67461b4ca1a9c2eb8ea1c0cc4bc39d91cf0d2823/odoo/http.py#L1456 Loading the modules starts a few lines below in https://github.com/odoo/odoo/blob/67461b4ca1a9c2eb8ea1c0cc4bc39d91cf0d2823/odoo/http.py#L1471 and after that, any request is correctly patched by

Yeah, you are right on this. Anyway I've never had such issue if not for 1st calls done after restart in a dev env where there's no other requests before. Which is rare. 
Probably loading at start is a good idea.

@lmignon ?

lmignon (approved):

SirTakobi:
I noticed that when there is no DB, the following occurred:
>   File "/path/to/odoo/src/odoo/odoo/http.py", line 1490, in dispatch
    result = _dispatch_nodb()
  File "/path/to/odoo/src/odoo/odoo/http.py", line 1460, in _dispatch_nodb
    func, arguments = self.nodb_routing_map.bind_to_environ(request.httprequest.environ).match()
  File "/path/to/odoo/src/odoo/odoo/tools/func.py", line 24, in __get__
    value = self.fget(obj)
  File "/path/to/odoo/src/odoo/odoo/http.py", line 1312, in nodb_routing_map
    return routing_map([''] + odoo.conf.server_wide_modules, True)
  File "/path/to/odoo/src/odoo/odoo/http.py", line 969, in routing_map
    members = inspect.getmembers(o, inspect.ismethod)
  File "/usr/lib/python3.6/inspect.py", line 342, in getmembers
    value = getattr(object, key)
  File "/path/to/odoo/lib/python3.6/site-packages/odoo/addons/base_rest/controllers/main.py", line 103, in collection
    return _PseudoCollection(self.collection_name, request.env)
  File "/path/to/odoo/lib/python3.6/site-packages/werkzeug/local.py", line 343, in __getattr__
    return getattr(self._get_current_object(), name)
  File "/path/to/odoo/src/odoo/odoo/http.py", line 258, in env
    self._env = odoo.api.Environment(self.cr, self.uid, self.context)
  File "/path/to/odoo/src/odoo/odoo/http.py", line 228, in cr
    raise RuntimeError('request not bound to a database')
RuntimeError: request not bound to a database

Because the members of RestController are evaluated in https://github.com/odoo/odoo/blob/67461b4ca1a9c2eb8ea1c0cc4bc39d91cf0d2823/odoo/http.py#L969
but since there is no DB, this call to `request.env` https://github.com/OCA/rest-framework/blob/c417c7eaf7aa7ee4b9c880d326936322365060b3/base_rest/controllers/main.py#L103
causes the above error.

The solution I chose is to exclude the generic RestController from the list of all controllers, so its members are not evaluated.
Looking forward to any feedback.

simahawk (approved):
LG. Would you mind forward porting when merged?

lmignon (approved):

PR #277, opened 2022-08-10
[FIX] 14.0 pydantic: generate schema using field name

PierrickBrun:
ref: https://pydantic-docs.helpmanual.io/usage/schema/ 

> The schema is generated by default using aliases as keys, but it can be generated using model property names instead by calling MainModel.schema/schema_json(by_alias=False).

Before:

generated schema uses aliases (odoo's internal naming) to name fields.

The schema does not correspond to what's effectively returned.

Capture d’écran du 2022-08-10 14-44-14

After:

generated schema uses the field's name, masking odoo's internal naming.

The schema corresponds to what's effectively returned.

Capture d’écran du 2022-08-10 14-44-38

### DIFF
--- a/base_rest_demo/tests/data/partner_pydantic_api.json
+++ b/base_rest_demo/tests/data/partner_pydantic_api.json
@@ -248,7 +248,7 @@
             "title": "Street2",
             "type": "string"
           },
-          "zip": {
+          "zip_code": {
             "title": "Zip",
             "type": "string"
           },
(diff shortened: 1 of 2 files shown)


petrus-v (approved):

rvalyi (approved):
good catch!

jeremi:
I second the need for this change. Should we also change it for the params? 

https://github.com/OCA/rest-framework/blob/6cf2dc90bd0f9a21f0385d3acc51273d8bbfb8f1/base_rest_pydantic/restapi.py#L56-L59

PierrickBrun:
> I second the need for this change. Should we also change it for the params?
> 
> https://github.com/OCA/rest-framework/blob/6cf2dc90bd0f9a21f0385d3acc51273d8bbfb8f1/base_rest_pydantic/restapi.py#L56-L59

I don't think so because I don't notice this problem on query parameters 

Capture d’écran du 2022-09-07 17-07-45

lmignon (approved):

PR #279, opened 2022-08-19
[FIX] base_rest_pydantic: forword service to PydanticModel.from_params 

petrus-v:
...while managing PydanticModelList.from_params 

To reproduce create a service use PydanticModelList in input param and call that service

### DIFF
--- a/base_rest_pydantic/tests/test_from_params.py
+++ b/base_rest_pydantic/tests/test_from_params.py
@@ -23,7 +23,12 @@ class TestPydantic(SavepointCase):
         self.Model1: BaseModel = Model1
 
     def _from_params(self, pydantic_cls: Type[BaseModel], params: dict, **kwargs):
-        restapi_pydantic = restapi.PydanticModel(pydantic_cls, **kwargs)
+        restapi_pydantic_cls = (
+            restapi.PydanticModelList
+            if isinstance(params, list)
+            else restapi.PydanticModel
+        )
+        restapi_pydantic = restapi_pydantic_cls(pydantic_cls, **kwargs)
         mock_service = mock.Mock()
         mock_service.env = self.env
         return restapi_pydantic.from_params(mock_service, params)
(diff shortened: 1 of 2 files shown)


petrus-v:
@lmignon I would be happy to get reviews from yon on this pydantic topic !

base_rest_pydantic/tests/test_from_params.py
```diff
@@ -22,8 +22,14 @@ class Model1(BaseModel):
 
         self.Model1: BaseModel = Model1
 
-    def _from_params(self, pydantic_cls: Type[BaseModel], params: dict, **kwargs):
-        restapi_pydantic = restapi.PydanticModel(pydantic_cls, **kwargs)
+    def _from_params(
+        self,
+        pydantic_cls: Type[BaseModel],
```
lmignon: @petrus-v Something is wrong here. The second arg is the service not a BaseModel type.

lmignon (changes requested):

base_rest_pydantic/tests/test_from_params.py
petrus-v: I don't get it :-/... may be we should rename this method in unit test that call the from_params method on the PydanticModel for the given BaseModel. 

Service is moked bellow before calling from_params !

Do I missed something ?

petrus-v: we could call it `_from_params_call` ?

lmignon: @petrus-v My bad, we are into the tests module....

Why do you need to pass the restapi class to the `_from_params' method. We don't need it. IMO we've 2 possibilities:
* `restapi_pydantic = restapi.PydanticModelList if isinstance(params, list) else restapi.PydanticModel`
* Add an explicit parameter to the method to specify that the params are a list...

petrus-v: I'm fine with your first suggestion, providing the restapi class as parameter was a way for me to explicitly test what I wanted to test here !

I'll change that !

base_rest_pydantic/tests/test_from_params.py
```diff
@@ -44,3 +49,15 @@ def test_from_params_missing_required_field(self):
...
         with self.assertRaisesRegex(UserError, msg):
             self._from_params(self.Model1, {"description": "Instance Description"})
+
+    def test_from_params_pydantic_model_list(self):
+        params = [
+            {"name": "Instance Name", "description": "Instance Description"},
+            {"name": "Instance Name 2", "description": "Instance Description 2"},
+        ]
+        instances = self._from_params(self.Model1, params)
+        restapi.PydanticModelList(self.Model1)
```
lmignon: this line is useless

petrus-v: sure :sweat_smile:

petrus-v: here you are !

lmignon (approved):
Thank you @petrus-v

PierrickBrun (approved):
Thanks

PR #287, opened 2022-09-01
[14.0][FIX] requirements.txt: Remove duplicate apispec requirement

carmenbianca:
pip install -r requirements.txt fails if there are double requirements.

Signed-off-by: Carmen Bianca Bakker <carmen@coopiteasy.be>

See also #286


sbidoul:
This should not be necessary with reasonably recent pip versions.

PR #293, opened 2022-09-30
[16.0][FIX] base_rest_pydantic: forword service to PydanticModel.from_params

lmignon:
in subclass PydanticModelList.from_params

To reproduce create a service use PydanticModelList in input param and call that service

forward port of #292


lmignon:
fast tracking since already approved in 14.0

PR #296, opened 2022-10-05
[IMP] rest_log: handle odoo.http.Response objects in logs

SilvioC2C:
Handle `odoo.http.Response` objects when creating logs.

### DIFF
--- a/rest_log/components/service.py
+++ b/rest_log/components/service.py
@@ -122,6 +122,8 @@ class BaseRESTService(AbstractComponent):
         if args:
             params = dict(params or {}, args=args)
 
+        params = self._log_call_sanitize_params(params)
+
         result = kw.get("result")
         error = kw.get("traceback")
         orig_exception = kw.get("orig_exception")
@@ -151,6 +153,11 @@ class BaseRESTService(AbstractComponent):
         if not values or enabled_states and values["state"] not in enabled_states:
             return
         return env["rest.log"].sudo().create(values)
+
+    def _log_call_sanitize_params(self, params):
+        if "password" in params:
+            params["password"] = "<redacted>"
+        return params
 
     def _db_logging_active(self, method_name):
         enabled = self._log_calls_in_db
--- a/rest_log/components/service.py
+++ b/rest_log/components/service.py
@@ -115,10 +115,7 @@ class BaseRESTService(AbstractComponent):
 
     def _log_call_in_db_values(self, _request, *args, params=None, **kw):
         httprequest = _request.httprequest
-        headers = dict(httprequest.headers)
-        for header_key in self._log_call_header_strip:
-            if header_key in headers:
-                headers[header_key] = "<redacted>"
+        headers = self._log_call_sanitize_headers(dict(httprequest.headers or []))
         params = dict(params or {})
         if args:
             params.update(args=args)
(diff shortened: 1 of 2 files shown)


simahawk:
> Handle `odoo.http.Response` objects when creating logs.

Can you please explain what you are trying to achieve?
Also, rebase and drop the change on README.md.

SilvioC2C:
@simahawk 

Some endpoints will not return a dict as expected by the ``_dispatch_with_db_logging`` method (and the other methods called within), but an object of type ``odoo.http.Response``.
When that happens, the instance fails to handle it while creating logs (ie: tries to call ``json.dumps()`` on such object, which is not allowed).
This patch should fix the issue.

jcoux (approved):
Maybe you can just add a comment to explain in case of returned an attachment (PDF by example), the service will return directly a response and not a dict.

rest_log/components/service.py
```diff
@@ -68,7 +68,7 @@ def _dispatch_with_db_logging(self, method_name, *args, params=None):
         log_entry = self._log_call_in_db(
             self.env, request, method_name, *args, params=params, result=result
         )
-        if log_entry:
+        if log_entry and isinstance(result, dict):
```
simahawk, line 71: this has been fixed in v14 already (I think we lost it in v13) https://github.com/OCA/rest-framework/commit/4e78558a1ec4fdd77df35017d2f09f2b68d337d9
Please cherry-pick this commit instead.

rest_log/components/service.py
```diff
@@ -123,6 +123,12 @@ def _log_call_in_db_values(self, _request, *args, params=None, **kw):
             params = dict(params or {}, args=args)
 
         result = kw.get("result")
+        if isinstance(result, Response):
+            status_code = result.status_code
+            result = {"status_code": status_code}
```
simahawk: Maybe content length and filename from the header if they are available?
Also, please, add a test for this :pray:

simahawk (changes requested):

simahawk:
> @simahawk
> 
> Some endpoints will not return a dict as expected by the `_dispatch_with_db_logging` method (and the other methods called within), but an object of type `odoo.http.Response`. When that happens, the instance fails to handle it while creating logs (ie: tries to call `json.dumps()` on such object, which is not allowed). This patch should fix the issue.

This should be the description of the commit :wink: 
You can also shorten the message by removing "in logs" which is redundant for "rest_log".

SilvioC2C:
@simahawk 
Added headers to `result`, also added unit tests.
Mind taking a look again? :sweat_smile:

rest_log/components/service.py
```diff
@@ -123,6 +123,15 @@ def _log_call_in_db_values(self, _request, *args, params=None, **kw):
             params = dict(params or {}, args=args)
 
         result = kw.get("result")
+        # NB: ``result`` might be an object of class ``odoo.http.Response``,
+        # for example when you try to download a file. In this case, we need to
+        # handle it properly, without the assumption that ``result`` is a dict.
+        if isinstance(result, Response):
+            status_code = result.status_code
+            result = {"status": status_code, "headers": dict(result.headers)}
```
simahawk: I think you must sanitize headers here like is done above. Maybe move it to a new method.

Can you also cherry-pick this fix? https://github.com/OCA/rest-framework/pull/301

SilvioC2C: @simahawk done

simahawk (approved):

PR #297, opened 2022-10-06
[14.0] base_rest: do not leak passwords in exception msg

simahawk:
You might have sensitive data in the request.
If you are using base_rest_auth_user_service you'll get a pwd there. By default, let's assume such keys such keys should never be leaked in logs.


SilvioC2C (approved):
LGTM

lmignon (approved):

PR #298, opened 2022-10-06
[14.0] rest_log: add sanitize params hook

simahawk:
Provide a hook to customize stored params.
At the same time, assume passwords must be always cleaned up.

Somehow related to #297

### DIFF
--- a/rest_log/components/service.py
+++ b/rest_log/components/service.py
@@ -122,6 +122,8 @@ class BaseRESTService(AbstractComponent):
         if args:
             params = dict(params or {}, args=args)
 
+        params = self._log_call_sanitize_params(params)
+
         result = kw.get("result")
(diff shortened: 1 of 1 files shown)


SilvioC2C (approved):
LGTM

Maybe you can import `BLACKLISTED_LOG_PARAMS` from `base_rest/http.py` and use it in `_log_call_sanitize_params()`? Or maybe create a tool in `base_rest` module directly to do this?

simahawk:
> LGTM
> 
> Maybe you can import `BLACKLISTED_LOG_PARAMS` from `base_rest/http.py` and use it in `_log_call_sanitize_params()`? Or maybe create a tool in `base_rest` module directly to do this?

At this stage I'm not sure they should share this.

mmequignon (approved):

PR #301, opened 2022-10-07
[15.0][FWD] rest_log: add sanitize params hook

simahawk:
Provide a hook to customize stored params.
At the same time, assume passwords must be always cleaned up.

FWD port of #298


simahawk:
TODO: fwd port other changes from v14

lmignon (approved):

lmignon:
Fast tracking since the initial PR is already merged.

PR #307, opened 2022-10-25
[CHG] pydantic: enclose list responses in dict

PierrickBrun:
for now PydanticModelList returns an array, this PR aims to change that.

Instead the response will be a dict composed of `data`, the array and `metadata`, a dict describing the array.

Here is explanations about why it would be better to return a dict instead of an array: https://softwareengineering.stackexchange.com/questions/286293/whats-the-best-way-to-return-an-array-as-a-response-in-a-restful-api 

That way we can extend `ListMetadata` how we please in order to, for instance, add pagination inside it.

Another side effect is to make pydantic's responses compatible with locomotive, as it always expects a dict..

Here is an example of code using this mechanism: https://github.com/OCA/helpdesk/pull/373


PierrickBrun:
@lmignon What do you think about this ?

base_rest_pydantic/restapi.py
```diff
@@ -154,10 +159,23 @@ def from_params(self, service, params):
 
     def to_response(self, service, result):
         self._do_validate(result, "output")
-        return [
+        json_res = {}
+        json_res["metadata"] = self.to_metadata_response(service, result["metadata"])
+        json_res["data"] = [
             super(PydanticModelList, self).to_response(service=service, result=r)
```
hparfr: I dislike when we super() and change the object hierarchy.

Can we do something like:
```suggestion
            self.to_data_response(service=service, result=r)
```

base_rest_pydantic/restapi.py
```diff
@@ -154,10 +159,23 @@ def from_params(self, service, params):
 
     def to_response(self, service, result):
         self._do_validate(result, "output")
-        return [
+        json_res = {}
+        json_res["metadata"] = self.to_metadata_response(service, result["metadata"])
+        json_res["data"] = [
```
lmignon: Why do you modify the structure of the response? We don't want to always return "metadata" and "data". Even if it's a list. We must keep the generic behavior. If you need such an approach for your services you are free to develop your own rest parameter class. Do you've try to use the generic types to define a generic model to encapsulate your response?

```python
from typing import Any, TypeVar, Generic, List

_T = TypeVar("_T", bound=Any)

from extendable_pydantic import ExtendableModelMeta

from pydantic import BaseModel


class SearchResponse(BaseModel,  Generic[_T], metaclass=ExtendableModelMeta):
    size: int = None
    data: List[_T]


 # to use it:
class Partner(BaseModel, metaclass=ExtendableModelMeta):
    name: str

results = SearchResponse[Partner](data=[Partner(name=1)], size=1)

.... 
```

Don't know if it's supported by pyndantic nor pydantic_extendable.... But would be nice and generic

PierrickBrun: That was my first try but I think I lost track and did something too complicated...

I reverted and pushed another solution

PierrickBrun:
I will rebase before puting this PR in non-draft mode

base_rest_pydantic/restapi.py
lmignon: @PierrickBrun If fixed the use of generics with pydantic and extendable... see https://github.com/lmignon/extendable-pydantic/pull/1 (At this stage it doesn't work with py < 3.8)

PR #308, opened 2022-11-10
[FIX] base_rest_datamodel: str version not supported anymore

paradoxxxzero:
This uses the packaging.version Version for openapi_version parameter since str is not supported anymore in apispec 6.0.1, see upstream issue: https://github.com/marshmallow-code/apispec/issues/810


paradoxxxzero:
Nevermind, I fixed it upstream

PR #313, opened 2022-12-04
[MIG] base_rest, base_rest_auth_api_key, base_rest_datamodel, base_rest_demo, base_rest_pydantic, datamodel, extendable: Migration to 16.0

StefanRijnhart:
Started out from https://github.com/OCA/rest-framework/pull/312. I assigned co-authorship of the main migration commit to @Nikul-OSI.

* Added warning to base_rest about the module being deprecated and removed @lmignon as maintainer (as per https://github.com/OCA/rest-framework/pull/313#issuecomment-1339407725)
* Removed adding [non existing modules to the Odoo registry](https://github.com/OCA/rest-framework/blob/15.0/base_rest/models/rest_service_registration.py#L93-L103) as it is harmful in Odoo 16. It makes Odoo halt while fetching [web assets](https://github.com/odoo/odoo/blob/5410b7c2/odoo/addons/base/models/ir_asset.py#L204) and while [fetching tests to run](https://github.com/odoo/odoo/blob/5410b7c2/odoo/tests/loader.py#L18). Removing this code does not _seem_ to break anything.
* Introduced a new `restapi` route type to trigger the dispatcher provided by `base_rest` as the new dispatcher registry is segmented by route `type`.
* In Odoo 16 the routing is encoded in an `original_routing` [attribute of the controller methods](https://github.com/odoo/odoo/blob/5410b7c2/odoo/http.py#L679) rather than a `routing` attribute. I've applied the `ROUTING_DECORATOR_ATTR` constant throughout the code.

PR #318, opened 2022-12-07
[MIG] rest_log: Migration to 16.0

aktiv-heli-kantawala:
* [ ] depends on #313


simahawk:
@heliaktiv thanks for your work.
There are some changes to port from v14 to v15 and then here on v16. Could you please have a look?

aktiv-heli-kantawala:
Updated already. Thank you

lmignon (changes requested):
The module is still defined as ''installable': False'.... and depends on #313

PR #323, opened 2023-01-03
[MIG] [16.0] base_graphql

rconjour:
[MIG] [16.0] base_graphql


paulius-sladkevicius:
As I'm keen to see it in the v16, could be worth to ping @sbidoul :-)

graphql_base/controllers/main.py
```diff
@@ -77,7 +61,7 @@ def _process_request(self, schema, data):
             return response
         except HttpQueryError as e:
             result = json_encode({"errors": [{"message": str(e)}]})
-            headers = dict(e.headers)
+            headers = dict()
```
sbidoul: Why this change?

.pre-commit-config.yaml
```diff
@@ -10,7 +10,6 @@ exclude: |
   ^base_rest_pydantic/|
   ^datamodel/|
   ^extendable/|
-  ^graphql_base/|
```
sbidoul, line 13: Could you do graphql_demo in this same PR, or apply https://github.com/OCA/maintainer-tools/wiki/Use-temporary-reference%28s%29-to-another-pull-request%28s%29 in #324 so we can test on runboat?

sbidoul (commented):
Thanks for working on this @rconjour !

sbidoul:
Hm, dear bot, why did you not notify me about this PR, since I'm a maintainer of the migrated module?

graphql_base/controllers/main.py
rconjour: The HttpQueryError does not contain any headers anymore

.pre-commit-config.yaml
rconjour: Done!

graphql_base/controllers/main.py
sbidoul: I still see it [here](https://github.com/graphql-python/graphql-server/blob/master/graphql_server/error.py#L18), though. And this does not seem to have changed recently.

sbidoul: Hm, it looks like `e.headers` can be `None`, so I suggest this:
```suggestion
            headers = dict(e.headers or {})
```

rconjour: Done!

sbidoul:
Thanks!

/ocabot merge patch

PR #325, opened 2023-01-19
[FIX] version in the manifest

arif-m150:
Update version number to 16


lmignon:
@arif-m150 Thank you for the PR. Unfortunately the migration process is a little more complex than changing the version into the manifest file. You can check and help to this process by reviewing the #313 PR.

Issue #290, opened 2022-09-26
Demo app does not work properly


MrTob:
Demo app OCA rest-framework does not work properly

## Module
base_rest_demo

## Describe the bug
When querying the table "res.partner" using "partner REST services", the API throws an INTERNAL SERVER ERROR.

## To Reproduce
**Odoo 15.0**:

Working Fine
![chrome_doyfZrYuTy](https://user-images.githubusercontent.com/30464573/192373841-ec03112e-0a83-4e72-bb86-4b8c75913e75.png)

![chrome_Qcdxhu8zd6](https://user-images.githubusercontent.com/30464573/192373908-77d362c0-387c-4c09-98bd-6138741e6135.png)
Not working because the user with this id is not active
![datagrip64_NTkd87CD6e](https://user-images.githubusercontent.com/30464573/192373980-9884e41d-6b1a-4309-804f-883fdf8e00ad.png)

**Expected behavior**
Suggested solution return only active users otherwise 404 NOT Found

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

PR #328, opened 2023-01-25
[14.0][FIX] base_rest_demo - odoo-addon-pydantic is used inside naive_orm_model class, but was missing in dependency

FrancoMaxime:
Fix missing dependency to odoo-addon-pydantic causing a runtime error when you want to use base_rest_demo.
Because the odoo-addon-pydantic is needed at this line:
https://github.com/OCA/rest-framework/blob/54e52e126931ea2e7e119f3d71c05bf5d68496ec/base_rest_demo/pydantic_models/naive_orm_model.py#L6

This fix comes from a discussion in the pull request of @StefanRijnhart  https://github.com/OCA/rest-framework/pull/313

And it should solve this issue: https://github.com/OCA/rest-framework/issues/290

### DIFF
--- a/base_rest_demo/__manifest__.py
+++ b/base_rest_demo/__manifest__.py
@@ -17,6 +17,7 @@
         "base_rest_pydantic",
         "component",
         "extendable",
+        "pydantic",
     ],
     "data": [],
     "demo": [],


lmignon (approved):
Thank you @FrancoMaxime

PR #340, opened 2023-06-07
[14.0] fastapi, extendable_fastapi: Backport from 16.0

lmignon:


lmignon:
@sbidoul does runboat run on python3.6 for odoo 14?  How can we manage an addon that requires python3.8 while runboat uses python 3.6?

PR #342, opened 2023-06-09
[16.0][IMP] fastapi: New type OdooFieldSelection

lmignon:
A python type to represent a selection field from odoo into pydantic models.

This type allows you to define a pydantic field with the same constraints
as a selection field from odoo.
The type is a string and the possible values are the ones defined in the
selection field from odoo. When used, the validation will check that the
value is one of the possible values from the selection field. It will also
extend the documentation of the field with the possible values and a description
based on the selections (value: Title) and the help of the field.

Usage:

To work, the type needs to be used with the `Field` class from pydantic. The
`odoo_model_name` and `odoo_field_name` need to be specified in the `extra`
attribute of the field. The `odoo_model_name` is the name of the odoo model
where the field is defined and the `odoo_field_name` is the name of the field
in the odoo model.

```python
        from odoo.addons.pydantic.typing import OdooFieldSelection
        from pydantic import BaseModel, Field

        class PartnerModel(BaseModel):
            type: OdooFieldSelection = Field(
                odoo_model_name="res.partner", odoo_field_name="type")


```

This new type declaration is here since we need access to the odoo's environment  to retrieve information from the field at schema generation and field value validation. When a model is used in the processing of a fastapi request I can use the ContextVar used to store the env from the werkzeug request. This context var is defined here. I would like to find a transversal way to solve this issue to make this field working with any pydantic model used into odoo at any time.... If you've some ideas they are welcome.

### DIFF
--- a/fastapi/typing.py
+++ b/fastapi/typing.py
@@ -1,0 +1,128 @@
+# Copyright 2023 ACSONE SA/NV
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from typing import Optional
+
+from odoo import fields
+
+from pydantic import errors
+from pydantic.fields import ModelField
+
+from .context import odoo_env_ctx
+
+
+class OdooFieldSelection(str):
+    """
+    A type to represent a selection field from odoo.
+
+    This type allows you to define a pydantic field with the same constraints
+    as a selection field from odoo.
+    The type is a string and the possible values are the ones defined in the
+    selection field from odoo. When used, the validation will check that the
+    value is one of the possible values from the selection field. It will also
+    extend the documentation of the field with the possible values and a description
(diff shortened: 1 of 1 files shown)


fastapi/typing.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        Modify the schema of the pydantic field to include the possible values
+        and the description of the odoo field into the OpenAPI schema.
+        """
+        if field:
+            selection = cls._get_selection(field)
+            field_schema["enum"] = selection
+            field_schema.pop("odoo_model_name", None)
+            field_schema.pop("odoo_field_name", None)
+            if not field_schema.get("desciption"):
```
FernandoRomera, line 124: Typo desciption => descRiption

PR #344, opened 2023-06-13
[16.0][FIX] Allow endpoint with slash in root path

flachica:
If I define a endpoint with slash in `Root path` this is not found and throw an 404 error

### DIFF
--- a/fastapi/fastapi_dispatcher.py
+++ b/fastapi/fastapi_dispatcher.py
@@ -20,11 +20,14 @@ class FastApiDispatcher(Dispatcher):
         # don't parse the httprequest let starlette parse the stream
         self.request.params = {}  # dict(self.request.get_http_params(), **args)
         environ = self._get_environ()
-        root_path = "/" + environ["PATH_INFO"].split("/")[1]
+        root_path = "/".join(environ["PATH_INFO"].split("/")[0:-1])
         # TODO store the env into contextvar to be used by the odoo_env
         # depends method
         fastapi_endpoint = self.request.env["fastapi.endpoint"].sudo()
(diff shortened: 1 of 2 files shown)


lmignon (changes requested):
@flachica IMO it's a bad idea since `/` as `Root path` could lead to collision with Odoo's routes. Your app should always have its specific root path as mount point into the Odoo server. I should enforce this rule on the `fastapi.endpoint` model.

flachica:
@lmignon First of all, thank you very much for your time.
Regarding the PR issue, I don't know if I'm understanding you. With my proposal I try that the "Root path" can contain the example values: "/app", "/app/v1" and "/app/v2". 

Currently it would only be possible to define "/app". With the proposed change the rest of the values would be allowed.

lmignon:
@flachica OK I understand now your point. The problem is wider. How can we deal with subpath into the root path at the same time of subpath in route? A route could also be declared as 'sales/items/'. I've to thing a little bit further to implement the right logic and support subpath for apps and routes.

simahawk:
> @flachica OK I understand now your point. The problem is wider. How can we deal with subpath into the root path at the same time of subpath in route? A route could also be declared as 'sales/items/'. I've to thing a little bit further to implement the right logic and support subpath for apps and routes.

@lmignon can you elaborate a bit on the issues you foresee?
In theory this should work out of the box. It's even documented here https://fastapi.tiangolo.com/reference/fastapi/ `app = FastAPI(root_path="/api/v1")`.

I haven't tried yet tho.

Also, depending on the use case you can use `prefix` in the routers to nest routes.

PR #345, opened 2023-06-14
[IMP] fastapi: Add fastapi addons directory structure into documentation

lmignon:
@sbidoul @AnizR @sebastienbeau @marielejeune what do you think about. I'll move the code from the demo app into the appropriate files if we agree on this.


lmignon:
Regarding the routers package, some tends to add a sub package with the version
```
routres
- v1
- - items
- v2
- - ..
```

fastapi/README.rst
```diff
@@ -1303,6 +1299,119 @@ for all the endpoints. You can do the same for a specific app by checking the
...
+related to the api. It will help you to easily find the files related to the api
+and it will help the other developers to understand your code.
+
+Here is the directory structure that we recommend. It's based on practices that
+are used in the python community when developing a fastapi app.
+
+.. code-block::
+
+  .
+  ├── x_api_addon
```
sbidoul:
```suggestion
  ├── x_api
```

fastapi/readme/USAGE.rst
```diff
@@ -1228,6 +1224,119 @@ for all the endpoints. You can do the same for a specific app by checking the
...
+* The **'models'** directory contains the odoo models. When you define a new
+  app, as for the others addons, you will add your new model inheriting from
+  the **'fastapi.endpoint'** model in this directory.
+* The **'routers'** directory contains the fastapi routers. You will add your
+  new routers in this directory. Each route starting with the same prefix should
+  be grouped in the same file. For example, all the routes starting with
+  '/items' should be defined in the **'items.py'** file. The **'__init__.py'**
+  file in this directory is used to import all the routers defined in the
+  directory and create a global router that can be used in an app. In
+  each router file, you will define a router without prefix and omit the
```
AnizR: ...and omit the _prefix_ ...

AnizR: or : ... and omit it ...

fastapi/depends.py
```diff
@@ -1,123 +1,12 @@
...
-from typing import TYPE_CHECKING
+import warnings
 
-from odoo.api import Environment
-from odoo.exceptions import AccessDenied
+warnings.warn(
+    "The 'depends' package is deprecated. Please use 'dependencies' instead.",
+    DeprecationWarning,
+    stacklevel=2,
+)
```
sbidoul, line 10: :heart:

fastapi/readme/USAGE.rst
```diff
@@ -1228,6 +1224,119 @@ for all the endpoints. You can do the same for a specific app by checking the
...
+  .. code-block:: python
+
+    from pydantic import BaseModel
+
+    class MyModel(BaseModel):
+        name: str
+        description: str = None
+
+  In the **'__init__.py'** file, you will import the model and add the prefix
+  to the model.
```
AnizR: I don't understand where is the "prefix" on the schema here in the example

AnizR (commented):
Great work!

fastapi/README.rst
```diff
@@ -1303,6 +1299,119 @@ for all the endpoints. You can do the same for a specific app by checking the
...
+  imported in the **'__init__.py'** file. This will allows to easily use
+  the same router in different apps with different prefixes. For example,
+  in your **'items.py'** file, you will define a router like this:
+  .. code-block:: python
+
+    router = APIRouter()
+
+    router.get("/", response_model=List[Item])
+
+    def list_items(..):
```
sbidoul:
```suggestion
    @router.get("/items", response_model=List[Item])
    def list_items(..):
```

I'd prefer to read the last part of the route close to where it is implemented.

In `base_rest` that was a source of confusion for me.

fastapi/README.rst
```diff
@@ -1303,6 +1299,119 @@ for all the endpoints. You can do the same for a specific app by checking the
...
+
+    router = APIRouter()
+    router.include_router(items_router, prefix="/items", tags=["items"])
+
+  When you include specific routers in the overall router, you should also
+  add the tags to the router. The tags will be used to group the routes in
+  the openapi documentation.
+* The **'schemas'** directory contains the pydantic models. You will add your
+  new pydantic models in this directory. Each model should be defined in a
+  separate file. The **'__init__.py'** file in this directory is used to import
```
sbidoul: separate file... not entirely sure I like that. I'd tend myself to write them all in `schemas/__init__.py` or `schemas.py` in simple cases, and start splitting them for big routers.

fastapi/README.rst
```diff
@@ -1303,6 +1299,119 @@ for all the endpoints. You can do the same for a specific app by checking the
...
+  In the **'__init__.py'** file, you will import the model and add the prefix
+  to the model.
+  .. code-block:: python
+
+    from .my_model import MyModel
+
+  This will allows to easily import models from the schemas package.
+  .. code-block:: python
+
+    from x_api_addon.schemas import MyModel
```
sbidoul, line 1404: :+1:

fastapi/README.rst
lmignon: may be `api_x` so you could also choose a specific prefix for dedicated projects like `shopinvade_api_x`

marielejeune: I think this is what we agreed on for shopinvader: `shopinvader_api_address`, `shopinvader_api_sale`...

fastapi/README.rst
lmignon: In such a case, it will never be possible to mount the router under another prefix (even if this will happen very rarely).  I improved the documentation on this point.

sbidoul: We can still mount by adding a prefix. But why would one want to change the last part of the route? If it's `/addresses`then it's fine to mount it at `/somewhere/addresses` but why would we mount the addresses end point at `/somethingentirelydifferent`? That would be very confusing, no?

fastapi/README.rst
```diff
@@ -1303,6 +1299,119 @@ for all the endpoints. You can do the same for a specific app by checking the
...
+  new routers in this directory. Each route starting with the same prefix should
+  be grouped in the same file. For example, all the routes starting with
+  '/items' should be defined in the **'items.py'** file. The **'__init__.py'**
+  file in this directory is used to import all the routers defined in the
+  directory and create a global router that can be used in an app. For example,
+  in your **'items.py'** file, you will define a router like this:
+
+  .. code-block:: python
+
+    router = APIRouter(tags=["items"]
```
AnizR: router = APIRouter(tags=["items"] **)** 
missing bracket

fastapi/README.rst
lmignon: Ok I simplified this point.

fastapi/readme/USAGE.rst
```diff
@@ -1228,6 +1224,119 @@ for all the endpoints. You can do the same for a specific app by checking the
...
+  '/items' should be defined in the **'items.py'** file. The **'__init__.py'**
+  file in this directory is used to import all the routers defined in the
+  directory and create a global router that can be used in an app. For example,
+  in your **'items.py'** file, you will define a router like this:
+
+  .. code-block:: python
+
+    router = APIRouter(tags=["items"])
+
+    router.get("items/", response_model=List[Item])
```
sbidoul:
```suggestion
    @router.get("/items", response_model=List[Item])
```

fastapi/readme/USAGE.rst
```diff
@@ -1228,6 +1224,119 @@ for all the endpoints. You can do the same for a specific app by checking the
...
+related to the api. It will help you to easily find the files related to the api
+and it will help the other developers to understand your code.
+
+Here is the directory structure that we recommend. It's based on practices that
+are used in the python community when developing a fastapi app.
+
+.. code-block::
+
+  .
+  ├── api_x
```
sbidoul: Actually, I liked to have `api` at the end. So the api modules are close to feature modules in alphabetical order.

lmignon: changed

fastapi/readme/USAGE.rst
lmignon: fixed

lmignon:
@sbidoul Is-it OK for you now? If it's, I'll adapt the demo code.

sbidoul:
Yes, good for me :)

PR #353, opened 2023-06-23
[WIP] fastapi: mount app at root_path instead of prefixing routes and docs URLs

sbidoul:
Since the FastAPI app for each endpoint is an independent app, mounted at a different prefix than the other, I'm wondering if the FastAPI mount mechanism would be more appropriate than prefixing all routers, prefixing the docs urls, etc.

I came to this while debugging the swagger oauth2 redirect url which is currently /docs instead of /{root_path}/docs.

When mounting the app at a prefix, FastAPI prefixes everything correctly.

This PR does not seem to work as is, but I wanted to ask if you had explored this approach @lmignon ?

### DIFF
--- a/fastapi/models/fastapi_endpoint.py
+++ b/fastapi/models/fastapi_endpoint.py
@@ -187,7 +187,9 @@ class FastapiEndpoint(models.Model):
         record = self.search([("root_path", "=", root_path)])
         if not record:
             return None
-        return ASGIMiddleware(record._get_app())
+        app = FastAPI()
+        app.mount(record.root_path, record._get_app())
+        return ASGIMiddleware(app)
 
     @api.model
     @tools.ormcache("root_path")
@@ -200,7 +202,7 @@ class FastapiEndpoint(models.Model):
     def _get_app(self) -> FastAPI:
         app = FastAPI(**self._prepare_fastapi_app_params())
         for router in self._get_fastapi_routers():
-            app.include_router(prefix=self.root_path, router=router)
+            app.include_router(router=router)
         app.dependency_overrides[dependencies.fastapi_endpoint_id] = partial(
             lambda a: a, self.id
         )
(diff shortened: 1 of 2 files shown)


fastapi/models/fastapi_endpoint.py
```diff
@@ -200,13 +200,15 @@ def get_uid(self, root_path):
...
-            app.include_router(prefix=self.root_path, router=router)
+            app.include_router(router=router)
         app.dependency_overrides[dependencies.fastapi_endpoint_id] = partial(
             lambda a: a, self.id
         )
         for exception, handler in self._get_app_exception_handlers().items():
             app.add_exception_handler(exception, handler)
-        return app
+        mainapp = FastAPI()
+        mainapp.mount(self.root_path, app)
```
lmignon: Not sure but I think that the path must be unique for each sub app you mount.

lmignon:
Sounds a good idea!

lmignon:
:champagne:

fastapi/models/fastapi_endpoint.py
sbidoul: I assume `root_path` is unique across endpoints

fastapi/tests/test_fastapi_demo.py
```diff
@@ -52,7 +54,7 @@ def tearDownClass(cls) -> None:
         super().tearDownClass()
 
     def _get_path(self, path) -> str:
-        return self.fastapi_demo_app.root_path + path
+        return path
```
sbidoul: But this shows it is a breaking change for tests that rely on `_get_app()` returning prefixed routes.

fastapi/tests/test_fastapi_demo.py
```diff
@@ -17,6 +18,7 @@
 from ..models.fastapi_endpoint_demo import EndpointAppInfo, ExceptionType
 
 
+@odoo.tests.tagged("post_install", "-at_install")
```
sbidoul, line 19: And I don't understand why this change suddenly requires tests to run post-install.

fastapi/tests/test_fastapi_demo.py
lmignon: But since we tests the routers and not the app in our addons, it's not a real issue.

fastapi/tests/test_fastapi_demo.py
lmignon: It was by chance...

lmignon (approved):
Works like a charm (Functional tests)

AnizR (approved):
Great, let's merge?

PR #359, opened 2023-07-10
[16.0][FIX] Make fastapi compatible with python 3.6

moitabenfdz:
Try to import Annotated from typing (python 3.9+), if import error import Annotated from typing_extensions

Change the pipe operator (|) (python 3.10+) to Union from typing_extensions

### DIFF
--- a/fastapi/README.rst
+++ b/fastapi/README.rst
@@ -203,7 +203,10 @@ that returns a list of partners.
 
 .. code-block:: python
 
-    from typing import Annotated
+    try:
+      from typing import Annotated
+    except ImportError:
+      from typing_extensions import Annotated
 
     from fastapi import APIRouter
     from pydantic import BaseModel
@@ -306,7 +309,10 @@ odoo models and the database from your route handlers.
 
 .. code-block:: python
 
-    from typing import Annotated
+    try:
+      from typing import Annotated
+    except ImportError:
+      from typing_extensions import Annotated
 
     from odoo.api import Environment
     from odoo.addons.fastapi.dependencies import odoo_env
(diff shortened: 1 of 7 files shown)


lmignon:
Thank you @moitabenfdz for your proposal. Nevertheless we don't want to support py36 for odoo 16. We start with py3.10.

moitabenfdz:
Thanks for your reply. My main concern is that doodba16 is in py3. 8.17, therefore, incompatible with fastapi module.  I found a TODO task in doodba pull requests from version 16.0 to switch to py3. 10 [#525](https://github.com/Tecnativa/doodba/pull/525) I'll be glad if @pedrobaeza  could give us some information about this, because I would like to use fastapi with doodba.

Thanks for your time, and have a good day!

pedrobaeza:
No work has been done yet to use P3.10 as base in Doodba for Odoo v16, as there hasn't been any blocker for now using P3.8. You can try to propose such Python version in Doodba and resolve the few problems related to packages/libraries (or maybe there has already been addressed). My doubt right now is which Python version is Odoo officially using for runbots v16 instances. We like to match with it, as it results battle-tested.

lmignon:
@moitabenfdz, @pedrobaeza 
* runboat and github actions uses 3.10 https://github.com/OCA/server-tools/actions/runs/5461843960/jobs/9940373783?pr=2670#step:4:54
* odoo runbot uses also 3.10 http://runbot122.odoo.com/runbot/static/build/47787223-saas-16-1/logs/install_base.txt

pedrobaeza:
Thanks for the info, Laurent. There's now a PoC to see the problems using that version in Doodba: https://github.com/Tecnativa/doodba/pull/553

PR #360, opened 2023-07-15
[16.0] Fastapi compatibility with pydantic v2

lmignon:


base_rest_pydantic/restapi.py
```diff
@@ -1,23 +1,12 @@
...
 
 from pydantic import BaseModel, ValidationError
-from pydantic.version import VERSION as PYDANTIC_VERSION
-
-PYDANTIC_V2 = PYDANTIC_VERSION.startswith("2.")
-
-if PYDANTIC_V2:
-
-    def validate_model(cls, data):
-        return cls.model_validate_json(data)
```
marielejeune: Can we remove `validate_model` as it only calls `model_validate_json`?

base_rest_pydantic/restapi.py
```diff
@@ -55,15 +45,20 @@ def from_params(self, service, params):
...
-        json_dict = result.dict()
-        if PYDANTIC_V2:
-            orm_mode = result.model_config.get("from_attributes", None)
-        else:
-            orm_mode = result.__config__.orm_mode
-        to_validate = json_dict if orm_mode else result.dict(by_alias=True)
-        *_, validation_error = validate_model(self._model_cls, to_validate)
-        if validation_error:
-            raise SystemError(_("Invalid Response %s") % validation_error)
+        json_dict = result.model_dump()
```
marielejeune, line 48: Not sure the name `json_dict` is still relevant as we need to convert it into json at line 54

fastapi/schemas.py
```diff
@@ -4,21 +4,20 @@
...
 
     total: int
     items: List[T]
 
 
 class Paging(BaseModel):
-    limit: Optional[int]
-    offset: Optional[int]
+    limit: Optional[int] = None
+    offset: Optional[int] = None
```
marielejeune, line 20: According to Pydantic migration guide:

_Pydantic V2 changes some of the logic for specifying whether a field annotated as Optional is required (i.e., has no default value) or not (i.e., has a default value of None or any other value of the corresponding type)._

Hence we must distinguish:
* `Optional[str]` (Required, can be `None`), and
* `Optional[str] = None` (Not required, can be `None`, is `None` by default).

Are we sure we want to make this update here?

See https://docs.pydantic.dev/latest/migration/#:~:text=Pydantic%20V2%20changes,annotations%20in%20V2%3A

marielejeune:
@lmignon ready for review

fastapi/schemas.py
lmignon: Yes. We must.

PR #363, opened 2023-07-26
[16.0][IMP] test_extendable_pydantic_fastapi: new test case

marielejeune:


marielejeune:
@lmignon New test case added

test_extendable_pydantic_fastapi/routers.py
```diff
@@ -21,11 +21,22 @@ def post_user(user: User) -> UserSearchResponse:
...
 
 @demo_pydantic_router.post("/post_private_user")
-def post_private_user(user: PrivateUser) -> UserSearchResponse:
+def post_private_user(user: PrivateUser) -> User:
     """A demo endpoint to test the extendable pydantic model integration
     with fastapi and odoo.
 
     Type of the request body is PrivateUser. This model inherits base model
-    User.
+    User but does not extend it.
```
lmignon, line 59: @marielejeune We should explain that for security reason, a method will always return attributes from declared response type. It should never return return attribute of a derived type from the declared response type even if in the route implementation we  return an instance of the derived type. The next route ensure that this assertion is also true with generics.

marielejeune: Done in a new commit

lmignon (approved):

PR #364, opened 2023-07-28
[16.0][ADD] New class StrictExtendableBaseModel into extendable_fastapi

marielejeune:
An `ExtendableBaseModel` (from `extendable_pydantic`) with strict validation.

### DIFF
--- a/extendable_fastapi/schemas.py
+++ b/extendable_fastapi/schemas.py
@@ -1,0 +1,29 @@
+from extendable_pydantic import ExtendableBaseModel
+
+
+class StrictExtendableBaseModel(
+    ExtendableBaseModel,
+    revalidate_instances="always",
+    validate_assignment=True,
+    extra="forbid",
+    strict=True,
+):
+    """
+    An ExtendableBaseModel with strict validation.
+
+    By default, Pydantic does not revalidate instances during validation, nor
+    when the data is changed. Validation only occurs when the model is created.
+    This is not suitable for a REST API, where the data is changed after the
+    model is created or the model is created from a partial set of data and
+    then updated with more data. This class enforces strict validation by
+    forcing the revalidation of instances when the method `model_validate` is
+    called and by ensuring that the values assignment is validated.
+
+    The following parameters are added:
+    * revalidate_instances="always": model instances are revalidated during validation
+    (default is "never")
+    * validate_assignment=True: revalidate the model when the data is changed (default is False)
+    * extra="forbid": Forbid any extra attributes (default is "ignore")
+    * strict=True: raise an error if a value's type does not match the field's type
+    annotation (default is False; Pydantic attempts to coerce values to the correct type)
+    """
--- a/extendable_fastapi/tests/test_strict_extendable_base_model.py
+++ b/extendable_fastapi/tests/test_strict_extendable_base_model.py
@@ -1,0 +1,68 @@
+import warnings
+from datetime import date
+
+from extendable_pydantic import ExtendableBaseModel
+
+from pydantic import ValidationError
+
+from ..schemas import StrictExtendableBaseModel
+from .common import FastAPITransactionCase
+
+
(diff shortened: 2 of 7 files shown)


lmignon (changes requested):
@marielejeune Thank you for this improvement. Can you pin the extendable_pydantic version to ">=1.1.0" into the manifest plz.

marielejeune:
@lmignon `extendable-pydantic` version updated + README.rst completed with `StrictExtendableBaseModel`

lmignon (approved):

extendable_fastapi/models.py
```diff
@@ -0,0 +1,20 @@
...
+
+class StrictExtendableBaseModel(
+    ExtendableBaseModel,
+    revalidate_instances="always",
+    validate_assignment=True,
+    extra="forbid",
+    strict=True,
+):
+    """
+    Add Config parameters to ExtendableBaseModel:
```
lmignon: Before explaining the how, it's worth explaining the what and the why ;-)

```suggestion
    An ExtendableBaseModel with strict validation.
    
    By default, Pydantic does not revalidate instances during validation, nor
    when the data is changed. Validation only occurs when the model is created.
    This is not suitable for a REST API, where the data is changed after the
    model is created or the model is created from a partial set of data and 
    then updated with more data. This class enforce strict validation by
    forcing the revalidation of instances when the method `model_validate` is
    called and by ensuring that the values assignment is validated.
    
    Add Config parameters to ExtendableBaseModel:
```

extendable_fastapi/models.py
```diff
@@ -0,0 +1,20 @@
+from extendable_pydantic import ExtendableBaseModel
```
lmignon: To follow the naming conventions of `odoo.addons.fastapi`, we should perhaps define this model in `schemas.py` :thinking:

extendable_fastapi/models.py
marielejeune: Added, thanks

extendable_fastapi/models.py
marielejeune: Done

extendable_fastapi/tests/test_strict_extendable_base_model.py
```diff
@@ -0,0 +1,64 @@
...
+    def test_StrictModel_revalidate_instance_always(self):
+        # Missing required fields and always revalidate
+        m = self.StrictModel.model_construct()
+        with self.assertRaises(ValidationError):
+            m.model_validate(m)
+
+    def test_Model_validate_assignment_false(self):
+        # Wrong assignment but no re-validation at assignment
+        m = self.Model(x=1, d=None)
+        m.x = "TEST"
```
marielejeune, line 34: This triggers a warning in the tests: 

```WARNING odoo py.warnings: /opt/odoo-venv/lib/python3.10/site-packages/pydantic/main.py:301: UserWarning: Pydantic serializer warnings:
  Expected `int` but got `str` - serialized value may not be as expected```
  
  This is coherent with the test case but we maybe want to avoid having this warning every time the tests are running?

**Review by lmignon — APPROVED** [review]

File: extendable_fastapi/schemas.py [hunk 4]
```diff
@@ -0,0 +1,29 @@
...
+    strict=True,
+):
+    """
+    An ExtendableBaseModel with strict validation.
+
+    By default, Pydantic does not revalidate instances during validation, nor
+    when the data is changed. Validation only occurs when the model is created.
+    This is not suitable for a REST API, where the data is changed after the
+    model is created or the model is created from a partial set of data and
+    then updated with more data. This class enforce strict validation by
```
AnizR:
```suggestion
    then updated with more data. This class enforces strict validation by
```

AnizR (approved):
One minor change that doesn't affect the behavior of the code.
Great work!

PR #365, opened 2023-07-31
[16.0][IMP] fastapi: add company_id dependency override

marielejeune:
The company can be specified on the `fastapi.endpoint`

### DIFF
--- a/fastapi/dependencies.py
+++ b/fastapi/dependencies.py
@@ -19,8 +19,20 @@ if TYPE_CHECKING:
     from .models.fastapi_endpoint import FastapiEndpoint
 
 
-def odoo_env() -> Environment:
-    yield odoo_env_ctx.get()
+def company_id() -> int | None:
+    """This method may be overriden by the FastAPI app to set the allowed company
+    in the Odoo env of the endpoint. By default, the company defined on the
+    endpoint record is used.
+    """
+    return None
+
+
+def odoo_env(company_id: Annotated[int | None, Depends(company_id)]) -> Environment:
+    env = odoo_env_ctx.get()
+    if company_id is not None:
+        env = env(context=dict(env.context, allowed_company_ids=[company_id]))
+
+    yield env
 
 
 def authenticated_partner_impl() -> Partner:
--- a/fastapi/models/fastapi_endpoint.py
+++ b/fastapi/models/fastapi_endpoint.py
@@ -48,6 +48,13 @@ class FastapiEndpoint(models.Model):
     docs_url: str = fields.Char(compute="_compute_urls")
     redoc_url: str = fields.Char(compute="_compute_urls")
     openapi_url: str = fields.Char(compute="_compute_urls")
+    company_id = fields.Many2one(
+        "res.company",
+        compute="_compute_company_id",
+        store=True,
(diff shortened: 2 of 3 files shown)


marielejeune:
@lmignon @sbidoul @AnizR Here is a new dependency override to set the company on the fastapi endpoint. 

I added this to https://github.com/shopinvader/odoo-shopinvader/pull/1364 to be able to play with it.

lmignon (approved):
Thank you @marielejeune LGTM.

fastapi/dependencies.py
```diff
@@ -19,8 +19,19 @@
     from .models.fastapi_endpoint import FastapiEndpoint
 
 
-def odoo_env() -> Environment:
-    yield odoo_env_ctx.get()
+def company_id() -> int | None:
+    """This method is overriden by the FastAPI app to make the company
+    available for your endpoint.
```
sbidoul:
```suggestion
    """This method may be overriden by the FastAPI app to define the allowed company
    in the Odoo environment of the endpoint. By default, the company defined on the endpoint record
    is used.
```

fastapi/dependencies.py
```diff
@@ -19,8 +19,20 @@
...
+    """This method may be overriden by the FastAPI app to make the company
+    available for your endpoint. By default, the company defined on the
+    endpoint record is used.
+    """
+    return None
+
+
+def odoo_env(company_id: Annotated[int | None, Depends(company_id)]) -> Environment:
+    env = odoo_env_ctx.get()
+    if company_id:
```
sbidoul:
```suggestion
    if company_id is not None:
```

Pas grave dans ce cas mais une bonne habitude à prendre (pour les situations où on veut distinguer 0 et None).

fastapi/dependencies.py
```diff
@@ -19,8 +19,20 @@
     from .models.fastapi_endpoint import FastapiEndpoint
 
 
-def odoo_env() -> Environment:
-    yield odoo_env_ctx.get()
+def company_id() -> int | None:
+    """This method may be overriden by the FastAPI app to make the company
+    available for your endpoint. By default, the company defined on the
+    endpoint record is used.
```
sbidoul:
```suggestion
    """This method may be overriden by the FastAPI app to set the allowed company
    in the Odoo env of the endpoint. By default, the company defined on the
    endpoint record is used.
```

AnizR (approved):
Great,Thanks :smile:

fastapi/models/fastapi_endpoint.py
```diff
@@ -85,6 +92,11 @@ def _compute_urls(self):
             rec.redoc_url = f"{rec.root_path}/redoc"
             rec.openapi_url = f"{rec.root_path}/openapi.json"
 
+    @api.depends("user_id")
+    def _compute_company_id(self):
+        for endpoint in self:
+            endpoint.company_id = endpoint.user_id.company_id
```
sbidoul, line 98: A little style remark: I think `for rec in self` or `for record in self`, when done consistently is easier to read than `for endpoint in self`. Just like self is... self, `rec` or `record`  for the current record in the self iterator.

PR #366, opened 2023-08-01
[16.0][REF] extendable_fastapi: unit tests

marielejeune:
* Move tests from `test_extendable_pydantic_fastapi` to `extendable_fastapi`
* Replace `from pydantic import BaseModel` with `from extendable_pydantic import ExtendableBaseModel`
* Add a unit test to ensure that with `extra='forbid'` config parameter in Pydantic models, a route cannot return a model with more attributes than what's expected in the response type.


lmignon (approved):
Thank you @marielejeune it makes sens to move these tests here since the test addon has the same dependencies as extendable_fastapi. I'll fast track this to clean the mess up introduces by the useless test addon

PR #368, opened 2023-08-14
[FIX] - base_rest: fix missing params

sbejaoui:


sbejaoui:
cc/ @lmignon

base_rest/components/service.py
```diff
@@ -151,7 +151,7 @@ def dispatch(self, method_name, *args, params=None):
         :return:
         """
         method = getattr(self, method_name, object())
-        params = params or {}
+        params = params or request.get_json_data() or {}
```
lmignon: @sbejaoui At this level, we should not use the `request` object. The parameters should be properly passed to the `dispatch` by callers. I've the feeling that the problem is upstream into the call stack.

lmignon:
closing since the fix is a workaround for a bug into [shopfloor_base](https://github.com/OCA/wms/pull/845)

PR #373, opened 2023-09-11
[FIX] fastapi: allow test with no authenticated partner

matthieusaison:
In case you need an anonymous call in your test.
The /fastapi/tests/common.py  _create_test_client() function fails due to an empty authenticated_partner_impl.

This proposal allows you to define an empty partner for your test:
 `cls.default_fastapi_authenticated_partner = cls.env["res.partner"]`
 
An authenticated test can be done like this: 
`self._create_test_client(partner=self.test_partner)`

### DIFF
--- a/fastapi/tests/common.py
+++ b/fastapi/tests/common.py
@@ -1,7 +1,6 @@
 # Copyright 2023 ACSONE SA/NV
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/LGPL).
 from contextlib import contextmanager
-from functools import partial
 from typing import Any, Callable, Dict
 
 from odoo.api import Environment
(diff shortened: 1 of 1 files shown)


sebastienbeau:
LGTM (code review), @lmignon ok for you ?

sebastienbeau (approved):

lmignon (changes requested):
The only change to do is: 

```python

partner = partner or self.default_fastapi_authenticated_partner or self.env["res.partner"]
dependencies[authenticated_partner_impl] = partial(lambda a: a, partner)


```

lmignon:
ping @matthieusaison

matthieusaison:
Thanks @lmignon for your feedback, the change is made.

lmignon (approved):
Code review

PR #374, opened 2023-09-13
[FIX] base_rest_pydantic: Fix json response validation

paradoxxxzero:
If there is non standard type in json (date for instance), `json.dumps` will crash.

This uses pydantic  `model_dump_json` method instead.

Not sure why the try catch is here though.


lmignon (approved):

FrancoMaxime (approved):
LGTM

rdualsam:
LGTM

PR #380, opened 2023-10-11
[16.0][IMP] extendable_fastapi: Add generics schemas

lmignon:
Add base schemas to be used to define pagination and paginated results when defining a search router.

### DIFF
--- a/extendable_fastapi/schemas.py
+++ b/extendable_fastapi/schemas.py
@@ -1,26 +1,19 @@
-from extendable_pydantic import ExtendableBaseModel
+from typing import Annotated, Generic, TypeVar
+
+from extendable_pydantic import StrictExtendableBaseModel
+
+from pydantic import Field
+
+T = TypeVar("T")
 
 
-class StrictExtendableBaseModel(
-    ExtendableBaseModel,
-    revalidate_instances="always",
-    validate_assignment=True,
-    extra="forbid",
-):
-    """
-    An ExtendableBaseModel with strict validation.
+class PagedCollection(StrictExtendableBaseModel, Generic[T]):
+    """A paged collection of items"""
 
-    By default, Pydantic does not revalidate instances during validation, nor
-    when the data is changed. Validation only occurs when the model is created.
-    This is not suitable for a REST API, where the data is changed after the
-    model is created or the model is created from a partial set of data and
-    then updated with more data. This class enforces strict validation by
-    forcing the revalidation of instances when the method `model_validate` is
-    called and by ensuring that the values assignment is validated.
+    # This is a generic model. The type of the items is defined by the generic type T.
+    # It provides you a common way to return a paged collection of items of
+    # extendable models. It's based on the StrictExtendableBaseModel to ensure
+    # a strict validation when used within the odoo fastapi framework.
 
-    The following parameters are added:
-    * revalidate_instances="always": model instances are revalidated during validation
-    (default is "never")
-    * validate_assignment=True: revalidate the model when the data is changed (default is False)
-    * extra="forbid": Forbid any extra attributes (default is "ignore")
-    """
+    count: Annotated[int, Field(..., description="The count of items into the system")]
+    items: list[T]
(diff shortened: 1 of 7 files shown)


extendable_fastapi/schemas.py
```diff
@@ -24,3 +26,28 @@ class StrictExtendableBaseModel(
     * validate_assignment=True: revalidate the model when the data is changed (default is False)
     * extra="forbid": Forbid any extra attributes (default is "ignore")
     """
+
+
```
lmignon, line 4: @sbidoul @sebastienbeau @qgroulard I'm not sure it's the right place to define these base schemas. Even if they are mainly used with fastapi, we also use the `StrictExtendableBaseModel` when we define models to be stored into a search engine. This leads to a useless dependency on fastapi. Do we've to create a specific addon for the `StrictExtendableBaseModel`? I could also move it to the `extendable_fastapi` python lib.

sbidoul: > I could also move it to the extendable_fastapi python lib.

But this PR *is* for `extendable_fastapi` ?

Question: why would one want to extend PagedCollection or Paging? Or is this necessary for compatibility when T is extendable?

lmignon: Yes, at least for PagedCollection, it's necessary for compatibility when T is extendable.

Regarding `StrictExtendableBaseModel`, it's not defined at the right place or we should not use-it when defining schemas for the search engine index.

sbidoul: Ah, yes I see. So you mean defining StrictExtendableBaseModel in `extendable_pydantic`, right?

lmignon: yes

sbidoul: Ok, makes sense to me. Let's do that.

sbidoul (approved):

extendable_fastapi/schemas.py
```diff
@@ -1,26 +1,17 @@
...
-    """
+class PagedCollection(StrictExtendableBaseModel, Generic[T]):
+    """A paged collection of items"""
+
+    # This is a generic model. The type of the items is defined by the generic type T.
+    # It provides you a common way to return a paged collection of items of
+    # extendable models. It's based on the StrictExtendableBaseModel to ensure
+    # a strict validation when used within the odoo fastapi framework.
+
+    total: int
```
sbidoul: Probably too late but... is `count` a better field name than `total`?

lmignon: IMO, It's not too late. What do you propose? `count` ?

sbidoul: Yes count, but it's not very important.

lmignon: Wording is important. I find a way to ensure the consistency with the same version of the model defined in the fastapi odoo addon but not supporting extendable items. The idea is to deprecated the use of the `total`field but keeping the compatibility on both side: json and python 

```python
import json
from typing import Annotated
from pydantic import BaseModel, computed_field, Field, AliasChoices


class Paginated(BaseModel):
    count: Annotated[
        int,
        Field(
            ...,
            desciption="Count of items into the system.\n Replaces the total field which is deprecated",
            validation_alias=AliasChoices("count", "total"),
        ),
    ]

    @computed_field()
    @property
    def total(self) -> int:
        return self.count

    @total.setter
    def total(self, value: int):
        self.count = value


p = Paginated(count=10)
print(p.model_dump_json())
>> {"count":10,"total":10}

p = Paginated(total=20)
print(p.model_dump_json())
>> {"count":20,"total":20}

p.total = 30
print(p.model_dump_json())
>> {"count":30,"total":30}

print(json.dumps(p.model_json_schema(), indent=2))
>> {
  "properties": {
    "count": {
      "desciption": "Count of items into the system.\n Replaces the total field which is deprecated",
      "title": "Count",
      "type": "integer"
    }
  },
  "required": [
    "count"
  ],
  "title": "Paginated",
  "type": "object"
}


```

lmignon: I've modified fastapi to ensure consistency between the two declarations in this template

lmignon:
@qgroulard @AnizR What do you think about this PR?

AnizR (approved):
Great!

qgroulard (approved):
Looks good.

sebastienbeau (approved):
LTGM, tested

PR #382, opened 2023-10-15
fastapi: add base service for exposing search/get

sebastienbeau:
Add generic service that can be used to expose odoo model


fastapi/services/fastapi_service_base.py
```diff
@@ -0,0 +1,51 @@
...
+        The purpose of the restrict domain is not restrict the information
+        dependending on the authenticated partner.
+        For security restriction, use security rule.
+        """
+        return []
+
+    def _convert_search_params_to_domain(self, params):
+        return []
+
+    def _search(self, paging, params=None):
```
sbidoul: In this case would it make sense to pass the domain instead of params?

sbidoul (commented):
In general I wonder if this would not be best implemented as helper functions rather than an abstract class.
This may lead to a more linear flow that could be easier to read.

lmignon (commented):
In my opinion, this approach lower the readability of the code. I would like to avoid to ends up with complex class inheritance where it becomes hard to know the responsibility of each layer and to understand the behavior of your final model.
You could implement the same functionality by defining for example a  specialized adapter class.

```python
from odoo import models
from odoo.osv import expressions
from extendable.main import ExtendableMeta


class FilteredDomainAdapter( metaclass=ExtendableMeta):

    def __init__(self, model: models.BaseModel, base_domain: list[tuple[]]):
        self._model = model
        self._base_domain = base_domain

    def get(self, record_id: int) -> BaseModel:
        return self._model.browse(record_id).filtered_domain(self._base_domain)

    def search_with_count(self, domain:list[tuple[], limit, offset]):
        final_domain = expressions.AND(
            [
                self._base_model,
               domain
            ]
        )
        ....

```

In your code (service or ...)

```python

sale_model = FilteredDomainAdapter(self.env["sale.order"],  [("type", "=", "delivery")])

my_order = sale_model.get(23)
count, orders = sale_model.search([("state", "in", ["cancel", "done"] )])

``` 

I'm not sure that `FilteredDomainAdapter` should be modifiable by inheritance. It should probably be a simple python class.

simahawk:
Well, I guess we have 2 topics here:

1. provide default endpoints to services
2. avoid duplicated logic for simple tasks

As for point 1, IMO is up to the service to explicitly declare its endpoints. You can define your own mixins if needed.

As for point 2, IMO having something like the `FilteredDomainAdapter` I see in the example makes it even harder/more complex than using components. If we have to provide such helper, I agree it can be a simple python class.

sebastienbeau:
Thanks for your feedback.
Indeed implementing a Helper can make it more readable and still avoid to duplicate the code.
I am going to test this approach.
Thanks

sebastienbeau:
Also regarding some good practice, I would like to have your feedback on this :

https://github.com/shopinvader/odoo-shopinvader/issues/1427
https://github.com/shopinvader/odoo-shopinvader/issues/1426

Thanks

sebastienbeau:
Hi @sbidoul @lmignon 
I confirm that the code is more readable see here : https://github.com/shopinvader/odoo-shopinvader/pull/1421/commits/169b339f7b16873a277c4dd86aff289e5ec752c1#diff-8661a778244e3a221d47f4c85ae9f1e05f275980e48ad3824616e3f10aab4556R65

fastapi/utils.py
```diff
@@ -0,0 +1,27 @@
+# Copyright 2023 Akretion (https://www.akretion.com).
```
lmignon, line 1: IMO this code has nothing to do with fastapi....

lmignon:
@sebastienbeau I close this one this the code is now provided into odoo-shopinvader

PR #387, opened 2023-10-31
[16.0] fastapi:  optionally authenticated partner

lmignon:


sebastienbeau (approved):
LGTM (code review)

simahawk (approved):
LG

qgroulard (approved):

lmignon:
@dreispt we should not merge this one....I should have put this PR in draft because I still want to check something. On further thought, the approach doesn't seem right. I'll revert it

PR #388, opened 2023-10-31
[16.0] fastapi: update schema to json_schema_extra

sbouillon:
PydanticDeprecatedSince20: Using extra keyword arguments on `Field` is deprecated and will be removed. Use `json_schema_extra` instead. (Extra keys: 'desciption'). Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.4/migration/

### DIFF
--- a/fastapi/schemas.py
+++ b/fastapi/schemas.py
@@ -11,13 +11,14 @@ T = TypeVar("T")
 
 class PagedCollection(BaseModel, Generic[T]):
     count: Annotated[
-        int,
+        int, 
         Field(
             ...,
-            desciption="Count of items into the system.\n "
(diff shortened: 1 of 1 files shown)


lmignon:
Hi @sbouillon 

Thank you for the proposal. Nevertheless this warning comes from a typo fixed [here](https://github.com/OCA/rest-framework/pull/387/commits/639bdf720f3ab49bf3553e00fde3a002770d13fa) Your fix is not the correct one. I propose to close your PR. Feels free to reopen it if you think I'm wrong.

Regards

sbouillon:
You're right. I wasn't looking closely enough :)

PR #389, opened 2023-11-03
[FIX] Corrected a Field parameter typo in schemas.py

meestaben:
Misspelt parameter was triggering a deprecation warning due to recent versions of Pydantic seeing it as an arbitrary parameter:

PydanticDeprecatedSince20: Using extra keyword arguments on `Field` is deprecated and will be removed. Use `json_schema_extra` instead. (Extra keys: 'desciption').

### DIFF
--- a/fastapi/schemas.py
+++ b/fastapi/schemas.py
@@ -14,7 +14,7 @@ class PagedCollection(BaseModel, Generic[T]):
         int,
         Field(
             ...,
-            desciption="Count of items into the system.\n "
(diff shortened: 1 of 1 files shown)


lmignon:
@meestaben Thank you for your fix. The same is also applied in #387. I will prepare a new releas with your commit and add a release note...

lmignon (approved):

lmignon:
Closing since your commit is included in #390.

Thank you for the contrib.

Issue #392, opened 2023-11-07
Extendable registry not cleaned up in case of error during the setup process


lmignon:
## Module
extendable_fastapi

## Describe the bug
When you override the `setUpClass` of `FastAPITransactionCase` in your own est class and an exception occurs in your override after the call to super, the registry used by the  `extendable` addon is not reset at its initial state.

**Expected behavior**
Once the test registry for `extendable` is  initialized in the `setUpClass` method of `FastAPITransactionCase`, we must ensure that any error that occurs into the code reset the registry at its initial state.

PR #393, opened 2023-11-07
[16.0][FIX] extendable_fastapi: Registry corruption in tests

lmignon:
Register the cleanup of the extendable registry just after its initialisation by calling the  method from unittest. If setUpClass() fails, meaning that tearDownClass() is not called, then any cleanup functions added with a call to  will still be called.

refs #392


sebastienbeau (approved):
LGTM (code review)

PR #394, opened 2023-11-07
[16.0] Update project template

lmignon:
* Uses ruff to check and format code
* Uses pyproject.toml files in addition to the setup directory


sbidoul:
Weird error. I'll look into this.

sbidoul:
It's a PYTHONPATH issue creating confusion because `import types` unexpectedly imports `types.py` at the root of the graphql_base addon. I'll investigate. It might be a bug in `build`, not sure.

sbidoul:
Should be fixed when https://github.com/OCA/oca-github-bot/pull/270 is deployed.

lmignon:
> Should be fixed when [OCA/oca-github-bot#270](https://github.com/OCA/oca-github-bot/pull/270) is deployed.

great! Thank you.

PR #398, opened 2023-11-14
17.0 mig graphql demo

dduarte-odoogap:


ccarrenoc (approved):

sbidoul:
Can you preserve the commit history using the recommended procedure? https://github.com/OCA/maintainer-tools/wiki/Migration-to-version-17.0

sbidoul:
I was wondering why this PR is green while the dependent `graphql_base` is not merged. But it does not set `graphql_demo` to installable, so that explains.

Could you cherry pick these commits in the `graphql_base` PR so we can test it on runboat?

dduarte-odoogap:
@sbidoul just pushed

PR #402, opened 2023-11-30
[16.0][FIX] fastapi_auth_jwt: unauthenticated partner

qgroulard:
Don't raise error if partner not found and unauthenticated partner is allowed

### DIFF
--- a/fastapi_auth_jwt/dependencies.py
+++ b/fastapi_auth_jwt/dependencies.py
@@ -204,8 +204,9 @@ class AuthJwtPartner(BaseAuthJwt):
         except Unauthorized as e:
             raise HTTPException(status_code=HTTP_401_UNAUTHORIZED) from e
         if not partner_id:
-            _logger.info("Could not determine partner from JWT payload.")
-            raise HTTPException(status_code=HTTP_401_UNAUTHORIZED)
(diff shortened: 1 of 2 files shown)


sbidoul:
Yes, this makes it do what the docs say it should do.

We need a test for that.

sbidoul (approved):

fastapi_auth_jwt/dependencies.py
```diff
@@ -204,8 +204,9 @@ def __call__(
         except Unauthorized as e:
             raise HTTPException(status_code=HTTP_401_UNAUTHORIZED) from e
         if not partner_id:
-            _logger.info("Could not determine partner from JWT payload.")
-            raise HTTPException(status_code=HTTP_401_UNAUTHORIZED)
+            if not self.allow_unauthenticated or validator.partner_id_required:
+                _logger.info("Could not determine partner from JWT payload.")
+                raise HTTPException(status_code=HTTP_401_UNAUTHORIZED)
```
sbidoul, line 209: Hm, a else is missing here.

sbidoul: Ok, no, this is correct because `browse(None)` works. I made #403 nevertheless.

PR #403, opened 2023-12-14
[FIX] fast_api_auth_jwt: improve readability

sbidoul:
Minore followup to #402 

Same result as before but more explicit.


sbidoul:
Fast tracking trivial change with no behaviour modification.

/ocabot merge patch

PR #404, opened 2023-12-15
16.0 add fastapi auth

sebastienbeau:
Introduce a new authentication based on partner, you can use it instead of external auth provider like keycloak

This module is linked to fastapi (it can maybe more generic in futur version).


simahawk:
@sebastienbeau If I'm not mistaken this work origins from https://github.com/akretion/rest-authenticable.
I don't get why it has been reworked as strictly tied to fastapi. IMO beside the routers this is something that should stay in a basic module as you did before.

BTW happy new year :)

sebastienbeau:
Yes it come from here.
Indeed we can split the module in two (generic, and specific to fastapi), I have hesitated
As base_rest is going to died and I do not see other case of using an alternative auth, I choose to depend on fastapi to avoid to split too much the code.
If we have case where the "base" code can be reuse I can resplit it.

simahawk:
> to avoid to split too much the code.

AFAIS 95% of the code would be in the standalone module, only a fraction goes into fastapi integration.

> If we have case where the "base" code can be reuse I can resplit it.

IMO is better to rollback to the other approach before you start spreading it :wink: 

In 4 years, when @lmignon will refactor these apis on the next api framework you'll thank yourself :stuck_out_tongue_winking_eye:

sebastienbeau:
Thanks for your feedback, we are going to re-split the module, @paradoxxxzero will work on it.

paradoxxxzero:
This PR depends on #438

lmignon:
I'll fix the error into the extendable-pydantic lib

lmignon:
@paradoxxxzero The compatibility issue is now fixed into extendable-pydantic... You've one failing test :smirk:

sebastienbeau:
Replaced by : https://github.com/OCA/rest-framework/pull/456

PR #410, opened 2024-01-12
[16.0][FIX] fastapi: always add authenticate_partner_id in context when evaluating ir rules

AnizR:
This fix is needed due to a modificiation of ir.rule that checks the domain when creating/modifying ir rules:

https://github.com/odoo/odoo/commit/91a356628e2878798bc2ff64af5c3fb3a6b3718c

The solution is to set authenticate_partner_id by default to False when it is not present in context

PR #411, opened 2024-02-09
[16.0][FIX] fastapi: add missing dependency

### LINKED ISSUE #367: [16.0][fastapi] "FastAPI Endpoint Runner" access rights causes login error
## Module
fastapi

## Describe the bug
Having "FastAPI Endpoint Runner" access rights causes an 500 Internal Server Error when trying to login.

## To Reproduce
**Affected versions**: 16.0

1. With fastapi installed, give an user "FastAPI Endpoint Runner" access rights.
2. Try to log in with that user.

**Expected behavior**
You can log in without any problem.

**Additional context**
Tested in runboat

TheMule71:
`security/ir_rule+acl.xml` refers to `authenticated_partner_id` which is defined in `base_rest`

Fixes: #367

### DIFF
--- a/fastapi/__manifest__.py
+++ b/fastapi/__manifest__.py
@@ -10,7 +10,10 @@
     "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
     "maintainers": ["lmignon"],
     "website": "https://github.com/OCA/rest-framework",
-    "depends": ["endpoint_route_handler"],
+    "depends": [
+        "base_rest",
+        "endpoint_route_handler",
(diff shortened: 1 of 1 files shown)


TheMule71:
This PR is a draft because I'm not entirely familiar with the code. Please comment.

It should be noted that the problem is rarely triggered.

lmignon:
The `authenticated_partner_id`doesn't come from `base_rest` but your issue has been fixed by  https://github.com/OCA/rest-framework/pull/410

PR #412, opened 2024-02-14
[14.0] [IMP] rest_log: add MissingError to dispatch method

renda-dev:
Add `RESTServiceMissingErrorException` to the exceptions to have the right status code returned when a `MissingError` is raised and intercepted


HekkiMelody (approved):
Code review

lmignon (approved):

PR #416, opened 2024-02-20
[16][FIX] fastapi: Compatibility with latest Odoo

lmignon:
From https://github.com/odoo/odoo/commit/cb1d057dcab28cb0b0487244ba99231ee292502e, the orignal werkzeug request is wrapped in a dedicated class to keep under control the attributes developers use. This change add code to make the fastapi addon working with version including this last change and prior version

fixes #414

PR #417, opened 2024-02-27
[14.0] [FIX] fastapi: Compatibility with latest Odoo

renda-dev:
Fixing tests for fastapi
- pin pydantic version in fastapi (not required for the repo, but mandatory for the module to work)
- update loading of middlewares according to https://github.com/encode/starlette/pull/2381
- <del>pin fastapi starlette version in extendable_fastapi (otherwise it would break the tests)</del>
- backport of https://github.com/OCA/rest-framework/pull/416


extendable_fastapi/__manifest__.py
```diff
@@ -16,6 +16,7 @@
     "external_dependencies": {
         "python": [
             "extendable-pydantic>=0.0.4",
+            "fastapi<=0.108.0",
```
lmignon: Why do you pin the fastapi version?

renda-dev: Actually I've done some more research and I've found out that [Starlette Version 0.35.0](https://github.com/encode/starlette/releases/tag/0.35.0) has some breaking changes that makes the tests fail.

I've previously pinned fastapi and thought it was working because `fastapi==0.108.0` has `starlette<0.33.0`

Please, let me know what you think! @lmignon

lmignon: @renda-dev Thank you for all your investigation! Could you explain what's wrong with starlette 0.35?

renda-dev: I've digged some more and found out that they changed the way that middlewares are declared and loaded, now it should be fixed with the latest commit

lmignon: Thank you!

HekkiMelody (approved):
Code review

lmignon (approved):

PR #419, opened 2024-03-13
[17.0] [FIX] Invalid version '15.0.1.0.1'. Modules should have a version in format `17.0.x.y.z`

rpinset:
After git clone this Odoo repo no longer wants to start because the version of the modules is not correct

```
 ValueError: Invalid version '15.0.1.0.1'. Modules should have a version in format `x.y`, `x.y.z`, `17.0.x.y` or `17.0.x.y.z`.
ValueError: Module base_rest_auth_user_service: invalid manifest
```

### DIFF
--- a/base_rest/__manifest__.py
+++ b/base_rest/__manifest__.py
@@ -6,7 +6,7 @@
     "summary": """
         Develop your own high level REST APIs for Odoo thanks to this addon.
         """,
-    "version": "16.0.1.0.2",
+    "version": "17.0.1.0.2",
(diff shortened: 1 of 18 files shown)


lmignon:
@rpinset Thank your for the proposal. A PR is submitted to odoo to fix this issue in a more generic way and restore the previous behavior for uninstallable addons. see https://github.com/odoo/odoo/pull/157657

sbidoul:
The upstream fix has been merged and should be included in the OCA CI image tomorrow.

PR #421, opened 2024-03-14
[16.0][FIX] fastapi: Handle error occurring outside the fastapi processing

lmignon:
To ensure that exception orrurring outside of the fastapi processing stack are properly handled, implement the 'handle_error' method into the FastApiDispatcher class

The `handle_error` method is the last one called by the WSGI app [when processing an exception](https://github.com/odoo/odoo/blob/16.0/odoo/http.py#L2070).... Therefore it must be implemented.

### DIFF
--- a/fastapi/fastapi_dispatcher.py
+++ b/fastapi/fastapi_dispatcher.py
@@ -4,6 +4,8 @@
 from contextlib import contextmanager
 from io import BytesIO
 
+from werkzeug.exceptions import InternalServerError
+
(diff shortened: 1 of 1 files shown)


lmignon:
ping @sbidoul @qgroulard @AnizR

AnizR (approved):
Code looks good to me :+1:
Great improvement, thanks

qgroulard (approved):
Code review

simahawk (approved):
makes sense

Issue #426, opened 2024-04-03
[16.0] fast_api_auth_jwt: AuthJwtOdooEnv: Incorrect Environment updated


jakobfischer93:
## Module
fast_api_auth_jwt

## Describe the bug
When using the `auth_jwt_authenticated_odoo_env` dependency, we get  the following error on routes using that dependency:
```
TypeError: odoo_env() got an unexpected keyword argument 'user'
```

## To Reproduce
**Affected versions**: 16.0

1. Create a route which has the dependency `auth_jwt_authenticated_odoo_env`
2. Access the route which uses that dependency

**Expected behavior**
We expected to get an odoo env with the user set according to the user id strategy.

**Additional context**
We were going to add a custom user id strategy. However, this issue also arises when using the default `static` user id strategy.
Seems to be the same issue as seen in #383.

sbidoul:
Thanks for reporting. Closing as duplicate of #406

Issue #383, opened 2023-10-25
Odoo 16.0, odoo_env doesn't have user as argument


usamahassan00015:
## Module
fastapi_auth_jwt,  fastapi

## Describe the bug
Error in odoo_env not having parameter "user"

_fastapi_auth_jwt/dependencies.py_
```
class AuthJwtOdooEnv(BaseAuthJwt):
    def __call__(
        self,
        request: Request,
        response: Response,
        authorization_header: Annotated[
            Optional[str],
            Depends(auth_jwt_http_header_authorization),
        ],
        default_validator_name: Annotated[
            Union[str, None],
            Depends(auth_jwt_default_validator_name),
        ],
        env: Annotated[
            Environment,
            Depends(odoo_env),
        ],
    ) -> Environment:
        validator = _get_auth_jwt_validator(
            self.validator_name or default_validator_name, env
        )
        payload, validator = _get_jwt_payload_and_validator(
            request, response, authorization_header, validator
        )
        uid = validator._get_and_check_uid(payload)
        return odoo_env(user=uid)
```
_fastapi_

```
def odoo_env(company_id: Annotated[int | None, Depends(company_id)]) -> Environment:
    env = odoo_env_ctx.get()
    if company_id is not None:
        env = env(context=dict(env.context, allowed_company_ids=[company_id]))

    yield env

```
## To Reproduce
 Create Route with AuthJwtOdooEnv function.

**Additional context**
OS Linux, Odoo 16.4, Python 3.10

tuanle96:
Hi, I found the solution to resolve your issue by replace with code below:

```python

class AuthJwtOdooEnv(BaseAuthJwt):
    def __call__(
        self,
        request: Request,
        response: Response,
        authorization_header: Annotated[
            Optional[str],
            Depends(auth_jwt_http_header_authorization),
        ],
        default_validator_name: Annotated[
            Union[str, None],
            Depends(auth_jwt_default_validator_name),
        ],
        env: Annotated[
            Environment,
            Depends(odoo_env),
        ],
    ) -> Environment:
        validator = _get_auth_jwt_validator(
            self.validator_name or default_validator_name, env
        )
        payload, validator = _get_jwt_payload_and_validator(
            request, response, authorization_header, validator
        )
        uid = validator._get_and_check_uid(payload)

        # return odoo_env(user=uid) #<- comment this line
        
        env.uid = uid 
        
        return env

```

usamahassan00015:
[...]

PR #427, opened 2024-04-03
FastAPI Auth JWT: Fixes incorrect environment access 

### LINKED ISSUE #406: fastapi_auth_jwt: Dependency auth_jwt_authenticated_odoo_env is broken
https://github.com/OCA/rest-framework/blob/880b7dc92eb487a8e11c24cd8698e9cdfd5f8931/fastapi_auth_jwt/dependencies.py#L239

`odoo_env` doesn't take argument `user`.

jakobfischer93:
It seems that the incorrect environment is used in the return of `AuthJwtOdooEnv`.
Instead of odoo_env (which ist no Environment, but a dependency, that only accepts a company_id), the actual Environment returned by the odoo_env dependency should be upgraded and returned.

Possible fix of #383 & #426.

closes #406

### DIFF
--- a/fastapi_auth_jwt/dependencies.py
+++ b/fastapi_auth_jwt/dependencies.py
@@ -236,7 +236,7 @@ class AuthJwtOdooEnv(BaseAuthJwt):
             request, response, authorization_header, validator
         )
         uid = validator._get_and_check_uid(payload)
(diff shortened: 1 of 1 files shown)


sbidoul (approved):
Looks good, thanks

PR #433, opened 2024-05-03
Revert "[16.0] fastapi:  optionally authenticated partner"

lmignon:
Reverts OCA/rest-framework#387

### DIFF
--- a/fastapi/dependencies.py
+++ b/fastapi/dependencies.py
@@ -43,29 +43,11 @@ def authenticated_partner_impl() -> Partner:
     See the fastapi_endpoint_demo for an example"""
 
 
-def optionally_authenticated_partner_impl() -> Partner | None:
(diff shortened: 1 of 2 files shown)


sbidoul:
@lmignon do you want to merge this according to https://github.com/OCA/rest-framework/pull/387#issuecomment-2092524062

lmignon:
@sbidoul I still have to think about it. The problem is not so trivial. Even if you declare the partner_id parameter as optional using type annotations, the implementation of the chosen method won't be aware of this annotation, which means that it will fail in the case of auth_jwt, for example, because it won't have found a token. Despite everything, I'm not comfortable with the proposed approach because I feel like I'm missing something. If I don't find another solution, the current code makes sense. If not, this PR could be merged with an explanation of how to cover this use case.

PR #442, opened 2024-07-09
[16.0] [FIX] fastapi: Unflag save_session in routing info

paradoxxxzero:
This is as much a PR as a question.

Is there any reason to not set the `save_session` to False and thus remove the `session_id` cookie? 

FYI: @sebastienbeau

### DIFF
--- a/fastapi/models/fastapi_endpoint.py
+++ b/fastapi/models/fastapi_endpoint.py
@@ -55,6 +55,16 @@ class FastapiEndpoint(models.Model):
         readonly=False,
         domain="[('user_ids', 'in', user_id)]",
     )
+    save_http_session = fields.Boolean(
+        string="Save HTTP Session",
+        help="Whether session should be saved into the session store. This is "
+        "required if for example you use the Odoo's authentication mechanism. "
+        "Oherwise chance are high that you don't need it and could turn off "
+        "this behaviour. Additionaly turning off this option will prevent useless "
+        "IO operation when storing and reading the session on the disk and prevent "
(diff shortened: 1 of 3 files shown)


lmignon:
> This is as much a PR as a question.
> 
> Is there any reason to not set the `save_session` to False and thus remove the `session_id` cookie?

by ignorance ...

lmignon (approved):
it makes sense

sbidoul:
Does this have backward compatibility implications?

paradoxxxzero:
Not that I know of, the session should not be used in the api.

flotho (approved):
Make sense,
LGTM

lmignon:
We could make it configurable on the endpoint.

sbidoul:
I think there are people who use regular Odoo users to authenticate REST APIs. There were PR to that end for base_rest in the past, IIRC. So if this influences such use cases (I'm not sure myself), it's probably necessary to make it configurable.

sebastienbeau:
@sbidoul @lmignon I think we can start by just add this default value. Then in your custom you can still inherit the method "_get_routing_info" to change the value and also to change the "auth" params to use native odoo authentification

lmignon:
> @sbidoul @lmignon I think we can start by just add this default value. Then in your custom you can still inherit the method "_get_routing_info" to change the value and also to change the "auth" params to use native odoo authentification

https://github.com/akretion/rest-framework/pull/7 :smirk:

paradoxxxzero:
Merged (and linted :p)

sbidoul (approved):

PR #443, opened 2024-08-08
[16.0][FIX] fastapi: paging - correctly set authorised minimal value in paging function

AnizR:
# Small fix

`fastapi.Query` accepts several parameters (`ge`,`gt`,`le`,`lt`) but not `gte`. 

I think that the intention was to set the minimal value to 1. Therefore, `ge` should be used.

### DIFF
--- a/fastapi/dependencies.py
+++ b/fastapi/dependencies.py
@@ -101,7 +101,7 @@ def optionally_authenticated_partner(
 
 
 def paging(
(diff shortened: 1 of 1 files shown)


sbidoul (approved):
gte is not event [documented](https://fastapi.tiangolo.com/reference/parameters/#fastapi.Query)

lmignon (approved):

PR #456, opened 2024-10-02
[16.0] [ADD] auth_partner, fastapi_auth_partner

paradoxxxzero:
This PR superseeds #404 and implements the separation between a base module `auth_partner` and the fastapi implementation `fastapi_auth_partner`.

This PR also introduces a stateless token for password reset, email validation and impersonation and some fixes and minor improvements.

This is still in testing, hence the draft status


auth_partner/models/auth_partner.py
```diff
@@ -0,0 +1,311 @@
...
+            "unique (directory_id, login)",
+            "Login must be uniq per directory !",
+        ),
+    ]
+
+    # hack to solve sql constraint
+    def _add_login_for_create(self, data):
+        partner = self.env["res.partner"].browse(data["partner_id"])
+        data["login"] = partner.email
+
```
sebastienbeau: This hack is no more needed in 16. You can use the option precompute=True on the field, so it will be computed before the create

fastapi_auth_partner/README.rst
```diff
@@ -0,0 +1,136 @@
...
+        def _get_fastapi_routers(self):
+          if self.app == "myapp":
+              return [
+                  auth_router,
+              ]
+          return super()._get_fastapi_routers()
+    
+        def _get_app_dependencies_overrides(self):
+            res = super()._get_app_dependencies_overrides()
+            if self.app == "portal":
```
lmignon:
```suggestion
            if self.app == "myapp":
```

auth_partner/models/auth_partner.py
```diff
@@ -0,0 +1,307 @@
...
+    def _request_reset_password(self):
+        return self.directory_id._send_mail_background(
+            "reset_password",
+            self,
+            callback_job=self.delayable()._on_reset_password_sent(),
+            token=self._generate_set_password_token(),
+        )
+
+    def _set_password(self, directory, token, password):
```
sebastienbeau, line 264: Can you add some comment here, (or in the decode_token method ?) regarding the key_salt ?
I know that you add this logic to make the token usable only one (as the function depend on the encrypted password)
But new contributor need help to understand this logic ?

Thanks

auth_partner/models/res_partner.py
```diff
@@ -0,0 +1,36 @@
...
+# License AGPL-3.0 or later (http://www.gnu.org/licenses/agpl).
+
+from odoo import fields, models
+
+
+class ResPartner(models.Model):
+    _inherit = "res.partner"
+
+    # guest logic should be moved in an dependency module
+    guest = fields.Boolean()
```
sebastienbeau: We can remove it, It's not needed anymore
```suggestion
```

lmignon:
@paradoxxxzero Can you take into account the last comments so we can move forward with your proposal.

paradoxxxzero:
> @paradoxxxzero Can you take into account the last comments so we can move forward with your proposal.

Done and squashed.

sebastienbeau (approved):
LGTM (code review) and tested

lmignon (approved):

PR #464, opened 2024-10-16
[16.0] [IMP] fastapi: Factorize error handling and  use it in tests with raise_server_exceptions=False

paradoxxxzero:
As of now using `raise_server_exceptions=False` leads to all exceptions to be returned as 500.
 
This PR aims to share the error handling between odoo FastAPIDispatcher and the starlette TestClient to get meaningful response in tests.

This also adds a log of 500 exceptions in testing which is very useful to get unhandled exceptions traceback while running the tests.

### DIFF
--- a/fastapi/error_handlers.py
+++ b/fastapi/error_handlers.py
@@ -1,14 +1,56 @@
 # Copyright 2022 ACSONE SA/NV
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/LGPL).
+from typing import Tuple
 
-
-from starlette.exceptions import WebSocketException
+from starlette import status
+from starlette.exceptions import HTTPException, WebSocketException
 from starlette.middleware.errors import ServerErrorMiddleware
 from starlette.middleware.exceptions import ExceptionMiddleware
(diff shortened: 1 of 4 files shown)


fastapi/routers/demo_router.py
```diff
@@ -4,6 +4,7 @@
 The demo router is a router that demonstrates how to use the fastapi
 integration with odoo.
 """
+
```
lmignon: @paradoxxxzero Can you avoid to modify this file just for cosmetic?

lmignon (changes requested):
Thank you @paradoxxxzero for this useful change. A little comment to reduce the diff and it'll be perfect.

paradoxxxzero:
> Thank you @paradoxxxzero for this useful change. A little comment to reduce the diff and it'll be perfect.

Done :)

lmignon (approved):

lmignon:
ping @AnizR

AnizR (approved):
Great! Thanks for your contribution.

Code LGTM

PR #468, opened 2024-11-06
[16.0] [ADD] fastapi_encrypted_errors

paradoxxxzero:
This module adds a "ref" field in the error response of FastAPI.
This field is an AES encrypted string that contains the error message / traceback.
This encrypted string can be decrypted using the endpoint decrypt error wizard.

Tests are failing for an unrelated change in endpoint_route_handler : https://github.com/OCA/web-api/commit/66cf0ba1d254474fc041c317c6d16f8519c85fc5

This PR depends on #500


extendable_fastapi/fastapi_dispatcher.py
```diff
@@ -3,15 +3,16 @@
...
-    FastApiDispatcher as BaseFastApiDispatcher,
-)
 
 from extendable import context
 
 
-class FastApiDispatcher(BaseFastApiDispatcher):
+# Inherit from last dispatcher of fastapi registered
+# This handles multiple overloaded dispatchers
+class FastApiDispatcher(_dispatchers["fastapi"]):
```
lmignon, line 15: This approach doesn't take into account that the module where your dispatcher is defined is installed or not into the database...

lmignon (commented):
It's a very interesting approach.

lmignon (approved):
Code review

lmignon:
@paradoxxxzero Can you rebase to launch the CI. Is this addon used in production?

paradoxxxzero:
Rebased, and yes it used in production but not very well tested since we don't make errors :innocent:

sebastienbeau (approved):
LGTM (code review), tested in production

PR #472, opened 2024-11-12
[IMP] fastapi: detail to have ValueError exception message

anikeenko-viktor:
Previously, detail contained only "Internal Server Error" in case of ValueError exception.
Now, it contains an actual exception message

### DIFF
--- a/fastapi/__manifest__.py
+++ b/fastapi/__manifest__.py
@@ -5,7 +5,7 @@
     "name": "Odoo FastAPI",
     "summary": """
         Odoo FastAPI endpoint""",
-    "version": "17.0.3.0.1",
+    "version": "17.0.4.0.0",
     "license": "LGPL-3",
     "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
     "maintainers": ["lmignon"],
--- a/fastapi/error_handlers.py
+++ b/fastapi/error_handlers.py
@@ -1,6 +1,8 @@
 # Copyright 2022 ACSONE SA/NV
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/LGPL).
 
+import traceback
+
 from starlette import status
 from starlette.exceptions import HTTPException, WebSocketException
 from starlette.middleware.errors import ServerErrorMiddleware
@@ -10,6 +12,7 @@ from starlette.websockets import WebSocket
 from werkzeug.exceptions import HTTPException as WerkzeugHTTPException
 
 from odoo.exceptions import AccessDenied, AccessError, MissingError, UserError
+from odoo.tools import config
 
 from fastapi import Request
 from fastapi.encoders import jsonable_encoder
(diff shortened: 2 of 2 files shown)


lmignon (changes requested):
IMO we must be very careful about the provided informations in case of error. We are conscious of the fact that no information is returned in the event of an error that is too generic, in order to prevent it from being used in the development of API attack scenarios by hackers. So I'm not in favor of this change, but let's see what others think.

anikeenko-viktor:
Realised that handling only `ValueError` is not enough in my case. I think we should have ability to see detailed error info somehow. Probably need to add some kind of DEBUG flag and show detailed error info only if it's provided (in staging, local servers for example)

lmignon:
> Realised that handling only `ValueError` is not enough in my case. I think we should have ability to see detailed error info somehow. Probably need to add some kind of DEBUG flag and show detailed error info only if it's provided (in staging, local servers for example)

Why not but errors and stacktrace are available into the log file.

fastapi/error_handlers.py
```diff
@@ -20,7 +22,10 @@
 def convert_exception_to_status_body(exc: Exception) -> tuple[int, dict]:
     body = {}
     status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
-    details = "Internal Server Error"
+    if os.getenv("FASTAPI_DEBUG") == "1":
+        details = repr(exc)
+    else:
+        details = "Internal Server Error"
```
lmignon:
```suggestion
    details = "Internal Server Error"
    if config.get_misc("fastapi", "dev_mode"):
        traceback = "".join(traceback.format_exception(*sys.exc_info()))
        details = f"{details}\n{traceback}"
```

you must also import the right modules...

```
import sys
import traceback

from odoo.tools.config import config
```

anikeenko-viktor: Thanks for the suggestion. Applied with slight changes.

fastapi/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "name": "Odoo FastAPI",
     "summary": """
         Odoo FastAPI endpoint""",
-    "version": "17.0.3.0.1",
+    "version": "17.0.4.0.0",
```
lmignon, line 8: The version will be changed at merge time. Can you remove this change plz.

fastapi/error_handlers.py
```diff
@@ -21,6 +24,9 @@ def convert_exception_to_status_body(exc: Exception) -> tuple[int, dict]:
     body = {}
     status_code = status.HTTP_500_INTERNAL_SERVER_ERROR
     details = "Internal Server Error"
+    if config.get_misc("fastapi", "dev_mode"):
```
lmignon, line 27: This config should be added into the documentation in a `readme/CONFIGURE.rst` (ex https://github.com/OCA/rest-framework/blob/16.0/base_rest/readme/CONFIGURE.rst)

PR #476, opened 2024-11-19
[16.0][FIX] fastapi: Disable exception logger when exception is expected

lmignon:


lmignon:
ouchhh the tests fail unexpectedly. I hope there's no change in the Odoo core that would affect error handling in the fastpi module... I'll have to analyze.

PR #425, opened 2024-03-26
[14.0][IMP] base_rest: show first-level menu to admins only

HekkiMelody:
Small self-explanatory PR.

Issue #470, opened 2024-11-07
[16.0] fastapi: Tests broken


paradoxxxzero:
FYI fastapi tests are now broken due to this merge : https://github.com/OCA/web-api/commit/66cf0ba1d254474fc041c317c6d16f8519c85fc5

See in this PR  #468 for instance (https://github.com/OCA/rest-framework/actions/runs/11707312885/job/32606402396?pr=468)

lmignon:
Thank you @paradoxxxzero for the report. I'll dig into this problem once I'll have some time.

sbidoul:
cc @grindtildeath as the author of the above mentioned commit in case this rings a bell

grindtildeath:
@sbidoul I haven't analyzed how the above mentioned commit affects other auth flows or v16.0 as my original commit was on v14.0. 

I tried my best to describe what was happening with OAuth activated (ie outdated cursor when validating the session due to isolation level) in the commit message, but cannot tell much more.

I guess @simahawk forward ported the commit to avoid having to spend as much time as we did to find it out, but without reproducing and an in depth analysis I cannot be of much help :disappointed_relieved:

paradoxxxzero:
As far I as remember, the tests fail because it expects no commits on an error request and the aforementioned commit introduces a new dedicated cursor (to avoid using the request's one) which does indeed a [commit](https://github.com/odoo/odoo/blob/16.0/odoo/sql_db.py#L176) when exiting the [context manager](https://github.com/OCA/web-api/commit/66cf0ba1d254474fc041c317c6d16f8519c85fc5#diff-16e4f0e22fe5170f12165f5318114e00ca5ecdbf397180244a567053b7f1c275R61)

paradoxxxzero:
Pushed a fix upstream https://github.com/OCA/web-api/pull/77

PR #477, opened 2024-11-25
[16.0][IMP]base_rest: show first-level menu to admins only

PicchiSeba:
fw-port of https://github.com/OCA/rest-framework/pull/425 to `16.0`

### DIFF
--- a/base_rest/README.rst
+++ b/base_rest/README.rst
@@ -46,7 +46,9 @@ systems, therefore increasing maintenance costs.
 Once developed, an `OpenApi <https://spec.openapis.org/oas/v3.0.3>`_ documentation
 is generated from the source code and available via a
 `Swagger UI <https://swagger.io/tools/swagger-ui/>`_ served by your odoo server
-at `https://my_odoo_server/api-docs`.
+at `https://my_odoo_server/api-docs`. There is also a convenience menu
+visible by users with permission "Administration: Settings", but the
+docs endpoint itself is public.
 
 **Table of contents**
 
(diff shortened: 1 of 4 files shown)


PicchiSeba:
@lmignon I'm trying to investigate the broken tests but they fail even on the `16.0` branch without any of my proposed changes. Do you have any idea why?

lmignon:
@PicchiSeba I don't understand the problem. On my laptop all the tests succeed. I need to analyze this problem in depth.

PicchiSeba:
@lmignon I don't know what I'm doing wrong, the tests for the `fastapi` module consistently fail, even when the repo is at commit c89c938.

I run the tests with `invoke test -m fastapi`
Should I set an environment variable somewhere?  Could it be a missing dependency? (I get no error for that though)

lmignon:
> @lmignon I don't know what I'm doing wrong, the tests for the `fastapi` module consistently fail, even when the repo is at commit [c89c938](https://github.com/OCA/rest-framework/commit/c89c938741920b5a3afa9cfc3f233e793733dfff).
> 
> I run the tests with `invoke test -m fastapi` Should I set an environment variable somewhere? Could it be a missing dependency? (I get no error for that though)

something may have changed in Odoo and as I don't have the latest version on my machine, I can't see it. Unfortunately, I don't have the time right now to analyze the problem further.

PicchiSeba:
Understandable, thanks for the answer regardless

PicchiSeba:
As stated in https://github.com/OCA/rest-framework/issues/470, tests were broken because of a change in upstream odoo

giarve:
I was thinking that maybe is it better to show it with developer mode enabled + admin settings?

Just a guess, but this PR still definitely improves the situation right now (shown to all, even non-technical users or normal employees).

PR #478, opened 2024-12-03
[16.0] [ADD] fastapi_dynamic_app

paradoxxxzero:
Here's a POC of a dynamic app management for fastapi. It's very alpha but it seems to work.


lmignon (commented):
interesting

PR #479, opened 2024-12-11
[17.0][FIX] fastapi: partner env for tests

oerp-odoo:
`partner` record leaks default env and even if you specify `default_fastapi_running_user` to be custom user, when running tests, running env will still have default (superuser), which makes testing outcomes very different than you might expect.

Setting `partner` to use env with specified user, makes sure it won't leak from there.

### DIFF
--- a/fastapi/tests/common.py
+++ b/fastapi/tests/common.py
@@ -127,6 +127,7 @@ class FastAPITransactionCase(TransactionCase):
             or self.default_fastapi_authenticated_partner
             or self.env["res.partner"]
         )
+        partner = partner.with_env(env)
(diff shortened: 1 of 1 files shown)


lmignon (approved):
Thank you  for the fix, it makes sense.

oerp-odoo:
@lmignon when you plan to merge it?

lmignon:
> @lmignon when you plan to merge it?

I would like to collect more feedback from others users. 

ping @paradoxxxzero @qgroulard @AnizR

paradoxxxzero:
Seems fair and would explain some false negative I got :+1:

Issue #484, opened 2025-01-07
[16.0] fastapi: daemon thread spawned by `a2wsgi.ASGIMiddleware`


sebalix:
## Module
`fastapi`

## Describe the bug
When [`<fastapi.endpoint>.get_app`](https://github.com/OCA/rest-framework/blob/16.0/fastapi/models/fastapi_endpoint.py#L209-L216) is called, it returns the FastAPI app wrapped in `a2wsgi.ASGIMiddleware`.

If the parameter `loop` is not set, this class is spawning a daemon thread automatically:
https://github.com/abersheeran/a2wsgi/blob/v1.10.8/a2wsgi/asgi.py#L130-L133

Is there an event loop we could re-use, or create a specific one (not daemonized) instead of having this automatically created one each time a request is dispatched on FastAPI app?

**Additional context**
We are reporting this because we spotted several daemon threads on Odoo (more than 30 on some of them), but we are still not sure if they could be linked to a performance issue (still digging). Note: we are running this Odoo instance in multi-threads (so we could have these daemon threads running forever as no worker is recycled).

sebalix:
Any insight @lmignon about it (as you wrote this part)?

lmignon:
@sebalix. The event loop is not created each time a request is dispatched. It's created the first time you call the [<fastapi.endpoint>.get_app](https://github.com/OCA/rest-framework/blob/16.0/fastapi/models/fastapi_endpoint.py#L204-L216)  method on your thread or process since the result is stored into thread local cache by db `@tools.ormcache("root_path")` In this way we ensure a proper isolation of the event loop by request processing whatever the running mode of odoo.

sebalix:
Thank you for your answer. We dug a bit more and sadly, a new event loop as well as a new daemon thread are created each time the ORM cache is cleared, leading to multiple event loops/threads running in parallel.
This behavior is probably mitigated with Odoo running with workers (related threads vanished once the worker is recycled), but in multi-threads mode these loops/threads persist.
Here is a piece of code we can run in Odoo shell that reproduces the issue:

```python
>>> import threading
>>> threading.active_count()
1
>>> threading.enumerate()
[<_MainThread(MainThread, started 139891588409152)>]

>>> app = env["fastapi.endpoint"].get_app("/endpoint")    # Spawn a deamon thread to run a newly created event loop
>>> threading.active_count()
2
>>> threading.enumerate()
[...]

PR #485, opened 2025-01-08
[16.0][FIX] fastapi: prevent the creation of multiple event loops/threads

sebalix:
Attempt to address issue #484

### DIFF
--- a/fastapi/models/fastapi_endpoint.py
+++ b/fastapi/models/fastapi_endpoint.py
@@ -1,7 +1,9 @@
 # Copyright 2022 ACSONE SA/NV
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/LGPL).
 
+import asyncio
 import logging
+import threading
 from functools import partial
 from itertools import chain
 from typing import Any, Callable, Dict, List, Tuple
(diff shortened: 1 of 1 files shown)


fastapi/models/fastapi_endpoint.py
```diff
@@ -18,6 +20,8 @@
 
 _logger = logging.getLogger(__name__)
 
+event_loop = asyncio.new_event_loop()
```
sebalix, line 23: Question here: do we want one event loop to handle multiple databases / endpoints? Or create one dynamically based on the DB(+endpoint)? Is there any benefit to do that?

lmignon: @sebalix What problem are you trying to solve here?   I don't think we can share the same event loop between requests. This is to ensure that the processing of requests made in parallel is isolated.  The only problem I can imagine if Odoo runs in multi-thread mode and not multi-process mode is that you end up with zombie threads for event loops that were created for deleted odoo threads.... In this case, we don't currently have a mechanism for managing the lifecycle of event loop threads (this isn't a problem in multi workers mode).

sebalix: I do not have the whole context of this `fastapi` module, so from what I see in #486 means we can share the same event loop/thread for different `root_path`, that was my main question (regarding performance).

Closing this one in favor of #486, thanks!

PR #486, opened 2025-01-08
[16.0][FIX] fastapi: Avoid zombie threads

lmignon:
Each time a fastapi app is created, a new event loop thread is created by the ASGIMiddleware. Unfortunately, every time the cache is cleared, a new app is created with a new even loop thread. This leads to an increase in the number of threads created to manage the asyncio event loop, even though many of them are no longer in use. To avoid this problem, the thread in charge of the event loop is now created only once per thread / process and the result is stored in the thread's local storage. If a new instance of an app needs to be created following a cache reset, this ensures that the same event loop is reused.

refs #484

### DIFF
--- a/fastapi/models/fastapi_endpoint.py
+++ b/fastapi/models/fastapi_endpoint.py
@@ -1,14 +1,11 @@
 # Copyright 2022 ACSONE SA/NV
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/LGPL).
 
-import asyncio
 import logging
-import threading
 from functools import partial
 from itertools import chain
 from typing import Any, Callable, Dict, List, Tuple
 
-from a2wsgi import ASGIMiddleware
 from starlette.middleware import Middleware
 from starlette.routing import Mount
 
@@ -17,28 +14,10 @@ from odoo import _, api, exceptions, fields, models, tools
 from fastapi import APIRouter, Depends, FastAPI
 
 from .. import dependencies
+from ..middleware import ASGIMiddleware
+from ..pools import fastapi_app_pool
 
 _logger = logging.getLogger(__name__)
-
-
-# Thread-local storage for event loops
-# Using a thread-local storage allows to have a dedicated event loop per thread
-# and avoid the need to create a new event loop for each request. It's also
-# compatible with the multi-worker mode of Odoo.
-_event_loop_storage = threading.local()
-
-
-def get_or_create_event_loop() -> asyncio.AbstractEventLoop:
-    """
-    Get or create a reusable event loop for the current thread.
-    """
-    if not hasattr(_event_loop_storage, "loop"):
-        loop = asyncio.new_event_loop()
-        loop_thread = threading.Thread(target=loop.run_forever, daemon=True)
-        loop_thread.start()
-        _event_loop_storage.loop = loop
-        _event_loop_storage.thread = loop_thread
-    return _event_loop_storage.loop
 
 
 class FastapiEndpoint(models.Model):
@@ -220,14 +199,9 @@ class FastapiEndpoint(models.Model):
         return f"{self._name}:{self.id}:{path}"
 
     def _reset_app(self):
-        self.get_app.clear_cache(self)
+        fastapi_app_pool.invalidate(self.root_path, self.env)
 
     @api.model
-    @tools.ormcache("root_path")
-    # TODO cache on thread local by db to enable to get 1 middelware by
-    # thread when odoo runs in multi threads mode and to allows invalidate
-    # specific entries in place og the overall cache as we have to do into
-    # the _rest_app method
     def get_app(self, root_path):
         record = self.search([("root_path", "=", root_path)])
         if not record:
@@ -235,8 +209,7 @@ class FastapiEndpoint(models.Model):
         app = FastAPI()
         app.mount(record.root_path, record._get_app())
         self._clear_fastapi_exception_handlers(app)
-        event_loop = get_or_create_event_loop()
-        return ASGIMiddleware(app, loop=event_loop)
+        return ASGIMiddleware(app)
 
     def _clear_fastapi_exception_handlers(self, app: FastAPI) -> None:
         """
--- a/fastapi/models/fastapi_endpoint.py
+++ b/fastapi/models/fastapi_endpoint.py
@@ -1,7 +1,9 @@
 # Copyright 2022 ACSONE SA/NV
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/LGPL).
 
+import asyncio
 import logging
+import threading
 from functools import partial
 from itertools import chain
 from typing import Any, Callable, Dict, List, Tuple
@@ -17,6 +19,26 @@ from fastapi import APIRouter, Depends, FastAPI
 from .. import dependencies
 
 _logger = logging.getLogger(__name__)
+
+
+# Thread-local storage for event loops
+# Using a thread-local storage allows to have a dedicated event loop per thread
+# and avoid the need to create a new event loop for each request. It's also
+# compatible with the multi-worker mode of Odoo.
+_event_loop_storage = threading.local()
+
+
+def get_or_create_event_loop() -> asyncio.AbstractEventLoop:
+    """
+    Get or create a reusable event loop for the current thread.
+    """
+    if not hasattr(_event_loop_storage, "loop"):
+        loop = asyncio.new_event_loop()
+        loop_thread = threading.Thread(target=loop.run_forever, daemon=True)
+        loop_thread.start()
+        _event_loop_storage.loop = loop
+        _event_loop_storage.thread = loop_thread
+    return _event_loop_storage.loop
 
 
 class FastapiEndpoint(models.Model):
@@ -213,7 +235,8 @@ class FastapiEndpoint(models.Model):
         app = FastAPI()
         app.mount(record.root_path, record._get_app())
         self._clear_fastapi_exception_handlers(app)
-        return ASGIMiddleware(app)
+        event_loop = get_or_create_event_loop()
+        return ASGIMiddleware(app, loop=event_loop)
 
     def _clear_fastapi_exception_handlers(self, app: FastAPI) -> None:
         """
--- a/fastapi/models/fastapi_endpoint.py
+++ b/fastapi/models/fastapi_endpoint.py
@@ -15,7 +15,6 @@ from fastapi import APIRouter, Depends, FastAPI
 
 from .. import dependencies
 from ..middleware import ASGIMiddleware
-from ..pools import fastapi_app_pool
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


lmignon:
ping @sebalix

sebalix (approved):

fastapi/models/fastapi_endpoint.py
```diff
@@ -19,6 +21,26 @@
 _logger = logging.getLogger(__name__)
 
 
+# Thread-local storage for event loops
+# Using a thread-local storage allows to have a dedicated event loop per thread
+# and avoid the need to create a new event loop for each request. It's also
+# compatible with the multi-worker mode of Odoo.
+_event_loop_storage = threading.local()
```
sebalix: <s>Indeed, having one loop per thread is mandatory, missed this in #485 :+1: </s>
EDIT: changing my mind, not working as expected, see below :arrow_down:

lmignon:
@sbidoul additional expert advice would be welcome :smirk:

fastapi/models/fastapi_endpoint.py
```diff
@@ -19,6 +21,26 @@
...
+def get_or_create_event_loop() -> asyncio.AbstractEventLoop:
+    """
+    Get or create a reusable event loop for the current thread.
+    """
+    if not hasattr(_event_loop_storage, "loop"):
+        loop = asyncio.new_event_loop()
+        loop_thread = threading.Thread(target=loop.run_forever, daemon=True)
+        loop_thread.start()
+        _event_loop_storage.loop = loop
+        _event_loop_storage.thread = loop_thread
```
sbidoul: Is it necessary to store the loop thread here?

sbidoul (dismissed):
Just a minor question, but otherwise this makes sense and looks good to me.

fastapi/models/fastapi_endpoint.py
sebalix: I still have doubt here: this will attach the loop to the current thread, which is OK in workers mode as the current thread is always the main thread of the worker, but in multi-threads, that would mean we will have one loop per spawned thread to handle the user's request, which is not the main thread. Am I wrong?

I try to test all of this locally.

sbidoul: Oh, is Odoo spawning a new thread for each request in multithread mode?

sebalix: What I did to reproduce this issue:

1. Here is Odoo running in multi-threads (2 exactly, forced with `ODOO_MAX_HTTP_THREADS` + cron threads disabled, and without queue_job neither), so we have the main thread+2 = 3:
image

2. Request done on a FastAPI endpoint (e.g. `curl http://localhost:8069/{root_path}/sales`), one daemon thread spawned to handle the loop as expected:
image

3. Beside in a Odoo shell, I reset the cache and relaunch a request on FastAPI endpoint:
    ```python
    >>> env.registry.clear_caches()
    >>> env.registry.signal_changes()
    ```
    `curl http://localhost:8069/{root_path}/sales`

4. Repeating step 3. several times: I get a new thread each time, so we are not re-using the loop on requests, while it was working as expected with #485 because having the loop created at the module level was done only once in the main thread, not for each subsequent thread created by Odoo to handle user's requests.

fastapi/models/fastapi_endpoint.py
sebalix: IMO no, we don't need it

fastapi/models/fastapi_endpoint.py
sbidoul: @sebalix does that mean sharing the same event loop across threads? Is that safe?

sebalix (commented):
Not approving based on my test here: https://github.com/OCA/rest-framework/pull/486/files#r1908260365

fastapi/models/fastapi_endpoint.py
sebalix: > Oh, is Odoo spawning a new thread for each request in multithread mode?

Yes: https://github.com/odoo/odoo/blob/18.0/odoo/service/server.py#L205-L217

> @sebalix does that mean sharing the same event loop across threads? Is that safe?

I don't know... :man_shrugging: maybe not something supported by default (reading https://www.neilbotelho.com/blog/multithreaded-async.html ).

And what if we attach the event loop to the current thread? Or one not with daemon=True? So we get one event loop created for each user's request, does it imply a lot of overhead? (and we maybe lose the benefit of FastAPI by doing so)

lmignon:
@sbidoul @sebalix 
Following the last comments, I propose to put in place a pool of event loop thread. The pool capacity will be set to the number of workers or to the max number of thread. In the same time, we can override the method  __call__ of the middelware to request the loop at the start of the request processing and release it at the end. In this way we will ensure a proper isolation of the event loop by request and the reuse of the pool in a safe way in multi process and multi thread mode....

Code will folllow

lmignon:
@sebalix @sbidoul I've implemented a new approach based on queue.Queue to manage a dynamic pool of event loops and the fastapi application cache. I'm probably still missing a bit of code for shutting down the threads used for the event pool when the server stops. I haven't yet figured out how to implement this, even though the method is available on the pool to stop all the threads.

fastapi/pools/event_loop.py
```diff
@@ -0,0 +1,68 @@
...
+        """
+        Get an event loop from the pool. If no event loop is available, create a new one.
+        """
+        try:
+            return self.pool.get_nowait()
+        except queue.Empty:
+            loop = asyncio.new_event_loop()
+            thread = threading.Thread(target=loop.run_forever, daemon=True)
+            thread.start()
+            return loop, thread
```
sbidoul, line 27: Should this not be put in the pool?

sbidoul: no

fastapi/pools/event_loop.py
```diff
@@ -0,0 +1,68 @@
...
+            self.pool.put((loop, thread))
+
+    def __get_event_loop_and_thread(
+        self,
+    ) -> tuple[asyncio.AbstractEventLoop, threading.Thread]:
+        """
+        Get an event loop from the pool. If no event loop is available, create a new one.
+        """
+        try:
+            return self.pool.get_nowait()
```
sbidoul, line 22: If max_size is set, should this not be a blocking `get()` instead ? Otherwise if the pool is empty, it will create a new thread and possibly block when putting threads back in the pool when it is full again?

sbidoul: Which is a theoretical concern since `EventLoopPool` does not seem to ever be initialized with a max_size. So perhaps remove support for max_size for the sake of simplicity, knowing that the native Odoo mechanism will limit the number of concurrent threads?

lmignon: I'll remove max size. It's useless and could conflict with the native Odoo mechanism.

lmignon: Moreever, the code below is a reliquat from a first implementation where my idea was to apply any limits set by Odoo to the pool. It was a very bad idea and far too complex.

fastapi/pools/fastapi_app.py
```diff
@@ -0,0 +1,52 @@
...
+class FastApiAppPool:
+    def __init__(self):
+        self._queue_by_db_by_root_path: dict[
+            str, dict[str, queue.Queue[FastAPI]]
+        ] = defaultdict(lambda: defaultdict(queue.Queue))
+
+    def __get_app(self, env: Environment, root_path: str) -> FastAPI:
+        db_name = env.cr.dbname
+        try:
+            return self._queue_by_db_by_root_path[db_name][root_path].get_nowait()
```
sbidoul: Is this double defaultdict initialization thread safe?

lmignon: Indeed, the literature says that `defaultdict` is not thread-safe when operations like key creation and modification occur concurrently from multiple threads.... This is now fixed in https://github.com/OCA/rest-framework/pull/486/commits/289868dd3b8b67104560cf2dfdc22671e38b2211

lmignon: @sbidoul Thinking again about the fact that `defautldict` is not thread safe when assigning a value and creating the default value if a key is missing, I'm not sure it's a real problem in this case The aim of the cache here is to provide a `FasAPI` application for processing a request while ensuring that only one application is used for a request at a time and avoiding the creation of a `FasAPI` app for each request (creating an app is a costly operation as it involves analysing all the routes defined in the app, etc.). The only risk/consequence related to the fact that the assignment to the 'defaultdict' is not thread safe is the possible loss of app created at system startup.... In my opinion, this is without consequence.  Even more, I think it's almost impossible for this to happen. The first operation when an app is requested from the cache is to create a queue and assign it to the cache if it doesn't already exist. It's only at the end of the use of this app that it's going to be put in the queue (at this time there is no more assignment to the dict, it's therefore thread safe). In the end I've come to the conclusion that in the worst case scenario we could have created a queue object for nothing but nothing more.  Therefore, the last commit seems to me overkill since introducing an external lock around a queue.Queue can negate many of the advantages of its built-in thread safety if every access to the queue is wrapped in the lock. This approach can lead to unnecessary contention and reduce the performance benefits of queue.Queue's fine-grained, internal synchronization. 
What do you think about?

sbidoul: > worst case scenario we could have created a queue object for nothing but nothing more

I agree. On the other hand, especially with free threaded python around the corner, I'm not that comfortable with multiple thread manipulating the same dictionary.

> introducing an external lock around a queue.Queue

It's not a lock on the queue, it's a lock on the creation of the queue, right?
That should be really cheap and negligible compared to everything else going on.

lmignon:
This PR works in production for the last 6 weeks. I plan to merge-it tomorrow.  ping @AnizR @qgroulard @sebalix

AnizR:
> This PR works in production for the last 6 weeks. I plan to merge-it tomorrow. ping @AnizR @qgroulard @sebalix

Thanks for this pr and the ping :smile: 
I'll integrate this into my projects.

Small interrogation: can these 'zombie' threads be an explanation of a slow but steady increase of memory usage?

lmignon:
> Small interrogation: can these 'zombie' threads be an explanation of a slow but steady increase of memory usage?

It could be, but it's not the element that led to the detection of this problem, nor even a factor considered in the diagnostic phase.

fastapi/pools/fastapi_app.py
```diff
@@ -0,0 +1,130 @@
...
+    def __get_pool(self, env: Environment, root_path: str) -> queue.Queue[FastAPI]:
+        db_name = env.cr.dbname
+        return self._queue_by_db_by_root_path[db_name][root_path]
+
+    def __get_app(self, env: Environment, root_path: str) -> FastAPI:
+        pool = self.__get_pool(env, root_path)
+        try:
+            return pool.get_nowait()
+        except queue.Empty:
+            env["fastapi.endpoint"].sudo()
```
AnizR: I think that this line is not necessary and can be removed, right?

AnizR (changes requested):
Here is my review, thanks for you PR.
I have just one remark.
Besides that, the code seems good to me :+1:

fastapi/pools/fastapi_app.py
lmignon: @AnizR I removed the useless line of code.

AnizR (approved):

PR #488, opened 2025-01-09
[16.0][fix] fastapi: init inner_exception properly

### LINKED ISSUE #487: [16.0] fastapi: object has no attribute inner_exception
We are seeing this stack trace in the logs:

```
  File "/odoo/src/odoo/http.py", line 2061, in __call__
    response = request._serve_db()
  File "/odoo/src/odoo/http.py", line 1649, in _serve_db
    return service_model.retrying(self._serve_ir_http, self.env)
  File "/odoo/src/odoo/service/model.py", line 133, in retrying
    result = func()
  File "/odoo/src/odoo/http.py", line 1676, in _serve_ir_http
    response = self.dispatcher.dispatch(rule.endpoint, args)
  File "/odoo/external-src/rest-framework/extendable_fastapi/fastapi_dispatcher.py", line 19, in dispatch
    return super().dispatch(endpoint, args)
  File "/odoo/external-src/rest-framework/fastapi/fastapi_dispatcher.py", line 34, in dispatch
    if self.inner_exception:
AttributeError: 'FastApiDispatcher' object has no attribute 'inner_exception'
```

This is with an up-to-date fastapi addon on 16.0

simahawk:
It looks to me that if the app does not yield anything for the response you might encour into an attribute error as the attribute won't be defined yet.

In general, the init method sounds like a better place for shared attributes.

Fixes #487

### DIFF
--- a/fastapi/fastapi_dispatcher.py
+++ b/fastapi/fastapi_dispatcher.py
@@ -13,6 +13,11 @@ from .error_handlers import convert_exception_to_status_body
 class FastApiDispatcher(Dispatcher):
     routing_type = "fastapi"
 
+    def __init__(self, request):
+        super().__init__(request)
+        # Store exception to later raise it in the dispatch method if needed
+        self.inner_exception = None
+
     @classmethod
     def is_compatible_with(cls, request):
         return True
@@ -47,7 +52,6 @@ class FastApiDispatcher(Dispatcher):
     def _make_response(self, status_mapping, headers_tuple, content):
         self.status = status_mapping[:3]
         self.headers = dict(headers_tuple)
-        self.inner_exception = None
         # in case of exception, the method asgi_done_callback of the
         # ASGIResponder will trigger an "a2wsgi.error" event with the exception
         # instance stored in a tuple with the type of the exception and the traceback.


gurneyalex (approved):

fastapi/fastapi_dispatcher.py
```diff
@@ -47,7 +52,6 @@ def handle_error(self, exc):
     def _make_response(self, status_mapping, headers_tuple, content):
         self.status = status_mapping[:3]
         self.headers = dict(headers_tuple)
-        self.inner_exception = None
```
lmignon, line 50: IMO you still need to reset the inner_exception each time you process a new response

lmignon (changes requested):

fastapi/fastapi_dispatcher.py
simahawk: @lmignon my understanding, but correct me if I'm wrong, is that the dispatcher will be re-initialized for each request. Now, if the assumption of the re-init per request is correct I wouldn't bother handling this differently.

Another point: even assuming you are correct, if you reset it for each iteration you might lose a previous exception. Hence, we should probably do a check if it is valued already and do not override it.

WDYT?

lmignon: @simahawk I think I was wrong :smirk:

lmignon (approved):

lmignon:
It would be nice to have a test reproducing this error...

PR #493, opened 2025-01-29
[IMP] base_rest_pydantic: compatibility pydantic v2 and pydantic v1

clementmbr:
Makes possible to use pydantic v1 features with pydantic v2 installed.

Note:

- not possible on base_rest_demo because module use both pydantic and extendable_pydantic
- trick not working with fastapi, but need to investigate why


petrus-v:
#551 could probably helps !

PR #499, opened 2025-02-25
[18.0][FIX] fastapi: Forwardport 16.0 pullrequest 486 - Avoid zombie threads

lembregtse:
This is a forward-port of https://github.com/OCA/rest-framework/pull/486.

I am aware that that PR is not yet finalized / approved, but we required the fix for version 18.0.

We have modified the caching system to be aligned with Odoo's new way of doing caches and refreshes. 

We will keep this PR update with the downstream PR.

### DIFF
--- a/fastapi/models/fastapi_endpoint.py
+++ b/fastapi/models/fastapi_endpoint.py
@@ -16,7 +16,6 @@ from fastapi import APIRouter, Depends, FastAPI
 
 from .. import dependencies
 from ..middleware import ASGIMiddleware
-from ..pools import fastapi_app_pool
 
 _logger = logging.getLogger(__name__)
 
@@ -122,10 +121,10 @@ class FastapiEndpoint(models.Model):
         return tuple(res)
 
     @api.model
-    def _routing_impacting_fields(self) -> tuple[str]:
+    def _routing_impacting_fields(self) -> Tuple[str, ...]:
         """The list of fields requiring to refresh the mount point of the pp
         into odoo if modified"""
-        return ("root_path",)
+        return ("root_path", "save_http_session")
 
     #
     # end of endpoint.route.sync.mixin methods implementation
@@ -200,7 +199,17 @@ class FastapiEndpoint(models.Model):
         return f"{self._name}:{self.id}:{path}"
 
     def _reset_app(self):
-        fastapi_app_pool.invalidate(self.root_path, self.env)
+        self._reset_app_cache_marker.clear_cache(self)
+
+    @tools.ormcache()
+    def _reset_app_cache_marker(self):
+        """This methos is used to get a way to mark the orm cache as dirty
+        when the app is reset. By marking the cache as dirty, the system
+        will signal to others instances that the cache is not up to date
+        and that they should invalidate their cache as well. This is required
+        to ensure that any change requiring a reset of the app is propagated
+        to all the running instances.
+        """
 
     @api.model
     def get_app(self, root_path):
--- a/fastapi/models/fastapi_endpoint.py
+++ b/fastapi/models/fastapi_endpoint.py
@@ -121,7 +121,7 @@ class FastapiEndpoint(models.Model):
         return tuple(res)
 
     @api.model
-    def _routing_impacting_fields(self) -> Tuple[str, ...]:
+    def _routing_impacting_fields(self) -> tuple[str, ...]:
         """The list of fields requiring to refresh the mount point of the pp
         into odoo if modified"""
         return ("root_path", "save_http_session")
@@ -199,17 +199,12 @@ class FastapiEndpoint(models.Model):
         return f"{self._name}:{self.id}:{path}"
 
     def _reset_app(self):
-        self._reset_app_cache_marker.clear_cache(self)
-
-    @tools.ormcache()
-    def _reset_app_cache_marker(self):
-        """This methos is used to get a way to mark the orm cache as dirty
-        when the app is reset. By marking the cache as dirty, the system
-        will signal to others instances that the cache is not up to date
-        and that they should invalidate their cache as well. This is required
-        to ensure that any change requiring a reset of the app is propagated
-        to all the running instances.
+        """When the app is reset we clear the cache, the system will signal to
+        others instances that the cache is not up to date and that they should
+        invalidate their cache as well. This is required to ensure that any change
+        requiring a reset of the app is propagated to all the running instances.
         """
+        self.env.registry.clear_cache()
 
     @api.model
     def get_app(self, root_path):
--- a/fastapi/models/fastapi_endpoint.py
+++ b/fastapi/models/fastapi_endpoint.py
@@ -1,8 +1,10 @@
 # Copyright 2022 ACSONE SA/NV
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/LGPL).
 
+import asyncio
 import logging
 from collections.abc import Callable
+import threading
 from functools import partial
 from itertools import chain
 from typing import Any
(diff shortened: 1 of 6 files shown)


lmignon (changes requested):
Thank you @lembregtse for the forward port. Nevertheless, can you preserve the authorship of the initial changes in 16.0. To preserve the author of the initial change, simply make a cherry pick of the PR commit on branch 16.0 and then make your changes in a new commit. This will also make it easy to see what adaptations have been made between the two versions. 
Out of curiosity, what problems did you encounter with the previous implementation and in what context?

lembregtse:
@lmignon Ofcourse, my bad, done. We are running Odoo containerized with gunicorn and ran into an issue with a customer where we use FastAPI for ETL data imports, where on certain objects/calls where the cache was refreshed zombie threads were being created. 

Considering we have a quite strict thread spawning limitation in those containers, after about 60 to 70 calls, the threads got exhausted and the FastAPI app would no longer accept new connections until the worker was recycled. While debuggin the issues we came across the PR for version 16.0.

lmignon:
> @lmignon Ofcourse, my bad, done. We are running Odoo containerized with gunicorn and ran into an issue with a customer where we use FastAPI for ETL data imports, where on certain objects/calls where the cache was refreshed zombie threads were being created.

Thank you for the explanation and your changes. Indeed this PR should solve your issue. We weren't affected by this one at acsone, because our instances always run in multi-worker mode and not in multi-thread mode. Solving this problem was an opportunity to improve the management of the eventpool and cache, which even if they weren't problematic in our case, were still not optimal. Still out of curiosity, what are the motivations for preferring gunicorn to serve Odoo instead of Odoo's multiworker runner? Don't you lose the mechanisms for managing the memory per worker and the maximum process execution time?
Kind regards.

lmignon (approved):
LGTM (Code review only)

fastapi/pools/fastapi_app.py
```diff
@@ -0,0 +1,136 @@
...
+    def __get_pool(self, env: Environment, root_path: str) -> queue.Queue[FastAPI]:
+        db_name = env.cr.dbname
+        return self._queue_by_db_by_root_path[db_name][root_path]
+
+    def __get_app(self, env: Environment, root_path: str) -> FastAPI:
+        pool = self.__get_pool(env, root_path)
+        try:
+            return pool.get_nowait()
+        except queue.Empty:
+            env["fastapi.endpoint"].sudo()
```
lmignon: Can you remove this line?

fastapi/pools/fastapi_app.py
```diff
@@ -0,0 +1,136 @@
...
+            return env["fastapi.endpoint"].sudo().get_app(root_path)
```
lmignon:
```suggestion
        except queue.Empty:
            return env["fastapi.endpoint"].sudo().get_app(root_path)
```

fastapi/pools/fastapi_app.py
```diff
@@ -0,0 +1,136 @@
...
+                )
+                self.invalidate(env)
+            self.set_cache_sequence(key, value)
+
+    def invalidate(self, env: Environment, root_path: str | None = None) -> None:
+        db_name = env.cr.dbname
+        if root_path:
+            self._queue_by_db_by_root_path[db_name][root_path] = queue.Queue()
+        elif db_name in self._queue_by_db_by_root_path:
+            del self._queue_by_db_by_root_path[db_name]
```
veryberry, line 135: @lmignon Why do you think it is sufficient to delete content of _queue_by_db_by_root_path (ASGIMiddleware) this way? Threads created before supposed to be cleaned by Python GC? I've done some test with Odoo cache invalidation and logging current threads number, the number is incremented by 1  after every "Caches invalidated" signalling. The threads remain in the process.

veryberry: Screenshot 2025-03-17 at 11 08 05

Screenshot 2025-03-17 at 11 10 03
the difference between 2 logs is 1 Odoo cache invalidation.

veryberry: @lmignon I think I've found the issue. seems like you should have redefined "__init__" in ASGIMiddleware
(https://github.com/abersheeran/a2wsgi/blob/master/a2wsgi/asgi.py#L130)
because every time new loop is created 

Screenshot 2025-03-17 at 11 54 33

lmignon: @veryberry The middleware used is defined [here](https://github.com/OCA/rest-framework/pull/499/files#diff-9681ad18c667fdc32a0df20eeb26ed638016005f61a5feef05266689f34d14c9R21) and extend https://github.com/abersheeran/a2wsgi/blob/master/a2wsgi/asgi.py#L130. The loop used comes from a poo and if one is available into the pool it will be reused.

veryberry: @lmignon the problem is that you have extended its `__call__` method
But actually the `__init__` is triggered first, and there loop is alway None, and new loop is created every time you call  `return ASGIMiddleware(app)`. 
I've proved it locally controlling threads after every Odoo cache invalidation.

lmignon: This PR is not mine, I can't modify it. A fix could be to also override the `__init__` method as

```python

class ASGIMiddleware(a2wsgi.ASGIMiddleware):
    def __init__(
        self,
        app: ASGIApp,
        wait_time: Optional[float] = None
    ) -> None:
        super().__init__(app, wait_time, False)

    def __call__(
        self, environ: Environ, start_response: StartResponse
    ) -> Iterable[bytes]:
        with event_loop_pool.get_event_loop() as loop:
            return ASGIResponder(self.app, loop)(environ, start_response)

```

lmignon: see https://github.com/OCA/rest-framework/pull/508/files

veryberry: Thanks
@lmignon just wanted you to be aware of this issue and know your opinion on it. In my case I've moved context manager usage from `__call__` to `__init__`. It helped.

lmignon: > In my case I've moved context manager usage from `__call__` to `__init__`. It helped.

????

without seeing the code, I find it hard to imagine that it would work correctly with the change you describes. It's important that the eventloop theard can only be used for one call at a time. That's why it's the call method that's overloaded, because once the context manager exit, the pool becomes available again for another odoo process/thread.

veryberry: @lmignon  would you be so kind to explain why it's important that the eventloop theard can only be used for one call at a time?

lmignon: Think of the event loop thread like a single train track. If two trains (calls) try to use the same track at the same time, they’ll collide, and everything will break. To keep things running smoothly, we let one train (call) complete its journey before allowing another one on the track.

fastapi/models/fastapi_endpoint.py
```diff
@@ -199,14 +199,19 @@ def _endpoint_registry_route_unique_key(self, routing: dict[str, Any]):
         return f"{self._name}:{self.id}:{path}"
 
     def _reset_app(self):
-        self.env.registry.clear_cache()
+        self._reset_app_cache_marker.clear_cache(self)
```
lmignon: see https://github.com/OCA/rest-framework/issues/510

lmignon (changes requested):

lmignon:
@lembregtse Do you plan to finalize your great work on this PR?

lembregtse:
I have been sick the last week, I’ll apply the requested changes when I’m up and running!

Sent from Outlook for iOS<https://aka.ms/o0ukef>
________________________________

lmignon:
> I have been sick the last week, I’ll apply the requested changes when I’m up and running! Sent from Outlook for iOS<https://aka.ms/o0ukef>

Thank you @lembregtse

lmignon:
@lembregtse  no news on this one ?

lembregtse:
On it right now! Thanks for the reminder ;-)

lembregtse:
@lmignon I've applied the first and last change requests. I'm not sure if anything needs to be changed based on your discussion with @veryberry. Let me know if anything else is required, thanks!

lmignon:
@lembregtse Thank you for your work. Could you also include the [commit](https://github.com/OCA/rest-framework/pull/499#event-17650166135) from #508 This one is also required to avoid zombie threads...

lmignon:
thank you for your hard work. I finalized your work in #532 by including the missing commit.

PR #501, opened 2025-02-25
[16.0] [ADD] fastapi_log

paradoxxxzero:
> This module allows an endpoint to activate full request logging in a database model.
> 
> It is useful to debug production issues or to monitor the usage of a specific endpoint.

This is not a migration of [rest_log](https://github.com/OCA/rest-framework/tree/16.0/rest_log), this module aims to be simpler, at least in the first iterations.

~~This PR depends on #500~~


simahawk:
Thanks for working on this. I wanted to do it since a while :sweat_smile: 

Early feedback...

The way I was planning to handle this was to have a base module to collect any kind of "request/response log".
This can be used by fastapi, base_rest, edi framework, endpoints and any custom implementation of request/response feature.

Would you consider splitting this part to a module like `request_log` or something like that?

paradoxxxzero:
@simahawk for now I don't want to overengineer this module. But feel free to split it when done. As a side note about that, is there any advantage to have all heterogeneous logs in the same model/views? Because aside of that, the amount of factored code would be low.

SirPyTech:
Hi @paradoxxxzero, thanks for this PR!
Are you still working on this?

paradoxxxzero:
Hello, no we have it deployed, I just forgot to remove the draft status, thanks.

fastapi_log/fastapi_dispatcher.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def dispatch(self, endpoint, args):
+        self.request.params = {}
+        environ = self._get_environ()
+        root_path = "/" + environ["PATH_INFO"].split("/")[1]
+        fastapi_endpoint = (
+            self.request.env["fastapi.endpoint"]
+            .sudo()
+            .search([("root_path", "=", root_path)])
+        )
```
PicchiSeba, line 29: **question**: what do you think about calling the `get_uid` method of `fastapi.endpoint`?

This way, if we need to override the method, the same endpoint will be matched

fastapi_log/models/fastapi_log.py
```diff
@@ -0,0 +1,227 @@
...
+    response_content_type = fields.Char(
+        compute="_compute_response_headers_derived", store=True
+    )
+    response_content_length = fields.Integer(
+        compute="_compute_response_headers_derived", store=True
+    )
+
+    def _headers_to_dict(self, headers):
+        try:
+            return {key.lower(): value for key, value in headers.items()}
```
PicchiSeba, line 74: **issue**: we should redact sentitive headers

https://github.com/OCA/rest-framework/blob/e0e48e3c555ccf4ce24b95655e1b1e1a6b59e247/rest_log/components/service.py#L130-L132

https://github.com/OCA/rest-framework/blob/e0e48e3c555ccf4ce24b95655e1b1e1a6b59e247/base_rest/http.py#L109-L111

PicchiSeba (changes requested):
Code review

fastapi_log/fastapi_dispatcher.py
SirPyTech: `get_uid` would return the user ID, could you please clarify how that could be used in this piece of code?

In https://github.com/akretion/rest-framework/pull/8 I have included your PR https://github.com/OCA/rest-framework/pull/515 and used the new `get_endpoint`, is that what you meant?

fastapi_log/models/fastapi_log.py
SirPyTech: Right, added this too in https://github.com/akretion/rest-framework/pull/8.

fastapi_log/fastapi_dispatcher.py
PicchiSeba: Yeah that's what I meant. Sorry for the confusion.

fastapi_log/fastapi_dispatcher.py
```diff
@@ -0,0 +1,76 @@
...
+                raise e
+            else:
+                try:
+                    log and log.log_response(response)
+                except Exception as e:
+                    _logger.warning("Failed to log response", exc_info=e)
+            finally:
+                if not tools.config["test_enable"]:
+                    try:
+                        cr.commit()  # pylint: disable=E8102
```
lmignon, line 70: Do you want to rollback the all the process if you encounter an error only with the cursor used for logging?

HekkiMelody:
Hello @paradoxxxzero
Could I ask you to take a look at https://github.com/akretion/rest-framework/pull/8 , we've done a lot of work on it and think it brings several useful improvements.

Thank you!

paradoxxxzero:
Superseeded by #554

PR #504, opened 2025-02-27
[16.0] [FIX] fastapi Headers may have multiple keys

paradoxxxzero:
As per [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) and [RFC9651](https://www.rfc-editor.org/rfc/rfc9651.html)

> Multiple message-header fields with the same field-name MAY be present in a message

As [werkzeug.wrappers.Response](https://werkzeug.palletsprojects.com/en/stable/wrappers/#werkzeug.wrappers.Response) expect a dict or a list of (key, value) there's no need to cast to dict


lmignon (approved):

sebastienbeau (approved):
LGTM (code review), validated in production

PR #505, opened 2025-03-04
[18.0] [MIG] migrate base_rest_pydantic

pierre-halleux:


lmignon:
@pierre-halleux Can you fix this PR plz?

PR #358, opened 2023-07-01
[14.0][FIX] base_rest : swagger multi-url

xavier-bouquiaux:
when using base_rest_auth_api_key and base_rest_auth_jwt on the same services -> problem at url level in swagger <br/>
fix by this commit <br/>

cherry-picked from aab4cc96c53f64558065b3915bc21061d571d1b0

PR #506, opened 2025-03-04
[18.0] [MIG] migrate base_rest

pierre-halleux:


simahawk:
Reminder to fwd port https://github.com/OCA/rest-framework/pull/358 (to be checked w/ oca-port).

base_rest/__init__.py
```diff
@@ -0,0 +1,10 @@
+import logging
+from . import models
+from . import components
+from . import http
+
+logging.getLogger(__file__).warning(
```
simahawk: @lmignon it's now clear that we are going to support this for a longer period of time. I don't think keeping logging this has much value if not polluting logs.
Would you mind if we move this warning to the README?

base_rest/components/service.py
```diff
@@ -0,0 +1,214 @@
...
+    # to the api, the werkzeug request interpret params as unicode
+    # so we have to convert params from string to float to int
+    if isinstance(val, int):
+        return val
+    if val:
+        return int(float(val))
+    return None
+
+
+def to_bool(val):
```
simahawk, line 31: this could be replaced w/ `odoo.tools.str2bool`

checklog-odoo.cfg
```diff
@@ -1,3 +1,4 @@
 [checklog-odoo]
 ignore=
     WARNING.* 0 failed, 0 error\(s\).*
+    WARNING.* base_rest is deprecated and not fully supported anymore.*
```
simahawk: to be dropped if the comment above is approved :)

simahawk (approved):
Tech review, functional review and manual tests :green_circle: 
Thank you!

@lmignon one remark about the deprecation :wink:

base_rest/__init__.py
lmignon: You're right. It's Ok for me.

simahawk:
@lmignon great, thanks! We can merge if good for you :wink:

lmignon:
> @lmignon great, thanks! We can merge if good for you 😉

for sure :grin:

PR #507, opened 2025-03-04
[18.0] [IMP] base_rest_pydantic: add anyof support

pierre-halleux:


lmignon:
@pierre-halleux Can you fix this PR plz?

PR #508, opened 2025-03-17
[16.0][FIX] fastapi: Ensure no default eventloop thread is created

lmignon:
Before this change, an eventloop thread was created on each instanciation of the ASGIMiddleware class. We want to avoid this behavior to avoid the creation of zombie threads

### DIFF
--- a/fastapi/middleware.py
+++ b/fastapi/middleware.py
@@ -13,12 +13,26 @@ from typing import Iterable
 
 import a2wsgi
 from a2wsgi.asgi import ASGIResponder
+from a2wsgi.asgi_typing import ASGIApp
 from a2wsgi.wsgi_typing import Environ, StartResponse
 
 from .pools import event_loop_pool
 
 
 class ASGIMiddleware(a2wsgi.ASGIMiddleware):
+    def __init__(
+        self,
+        app: ASGIApp,
+        wait_time: float | None = None,
+    ) -> None:
+        # We don't want to use the default event loop policy
+        # because we want to manage the event loop ourselves
+        # using the event loop pool.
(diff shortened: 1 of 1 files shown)


lmignon:
ping @sebalix @AnizR

fastapi/middleware.py
```diff
@@ -13,12 +13,26 @@
...
+        app: ASGIApp,
+        wait_time: float | None = None,
+    ) -> None:
+        # We don't want to use the default event loop policy
+        # because we want to manage the event loop ourselves
+        # using the event loop pool.
+        # Since the the base class check if the given loop is
+        # None, we can pass False to avoid the initialization
+        # of the default event loop
+        super().__init__(app, wait_time, False)
```
AnizR, line 34: A bit hacky but I don't see any other option.

fastapi/middleware.py
```diff
@@ -13,12 +13,26 @@
...
 from a2wsgi.wsgi_typing import Environ, StartResponse
 
 from .pools import event_loop_pool
 
 
 class ASGIMiddleware(a2wsgi.ASGIMiddleware):
+    def __init__(
+        self,
+        app: ASGIApp,
+        wait_time: float | None = None,
```
AnizR, line 26: Shouldn't we need to keep the parent signature?

`loop: Optional[asyncio.AbstractEventLoop] = None,`

lmignon: IMO no. This class is designed to be used in the specific context of the fastapi addon where we enforce the use of the pool of eventloop threads.

AnizR (approved):

PR #509, opened 2025-03-18
[18.0][MIG] rest_log

glitchov:
Migration from 16.0


lmignon:
@glitchov Thank you for the migration. Can you fix the test plz?

lmignon:
ping @glitchov 
Could you also include #520

PR #513, opened 2025-03-25
[14.0][FIX] base_rest_pydantic: fix `_` being overridden in the `to_response` function

alexandregaldeano:
A `TypeError: 'list' object is not callable` error was raised when there was a `validation_error`. 
This is due to the fact that `_` was reassigned as a list and then we tried to called `_("Invalid Response %s")`.

### DIFF
--- a/base_rest_pydantic/restapi.py
+++ b/base_rest_pydantic/restapi.py
@@ -43,12 +43,13 @@ class PydanticModel(restapi.RestMethodParam):
             raise UserError(_("BadRequest %s") % ve.json(indent=0))
 
     def to_response(self, service, result):
-        # do we really need to validate the instance????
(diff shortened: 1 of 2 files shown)


nico-foodles (approved):
LGTM

petrus-v (approved):
Thanks for the fix.

We could probably add a unit test on this specific case to ensure behaviour persist!

damdam-s (approved):

petrus-v (approved):
thanks for the unit test addition :heart:

alexandregaldeano:
@petrus-v done, I also updated the comment within the `to_response` to explain why validating the json is necessary

lmignon (approved):

Issue #446, opened 2024-08-21
fastapi: root_path cannot contain a /


sbidoul:
I'd like to be able to use a root path such as /myapi/v1

This does not work because the fastapi dispatcher takes the first component of the url as the root path to search the endpoint

https://github.com/OCA/rest-framework/blob/2380e306e896c732ce20c66000f5bc094731920b/fastapi/fastapi_dispatcher.py#L32

oerp-odoo:
Yeah I was confused why my endpoint was not working as this is not documented in odoo fastapi (or maybe I did not see it?).

`/api/demo` gives internal server error, and `/api_demo` works.

PR #515, opened 2025-04-01
[16.0][IMP]fastapi: enable multi-slash routes

PicchiSeba:
This is a Work In Progress PR, which attempts to address this issue https://github.com/OCA/rest-framework/issues/446.
Any suggestion is welcome

EDIT: depends on https://github.com/OCA/rest-framework/pull/524

### DIFF
--- a/fastapi/models/fastapi_endpoint.py
+++ b/fastapi/models/fastapi_endpoint.py
@@ -198,6 +198,8 @@ class FastapiEndpoint(models.Model):
         return f"{self._name}:{self.id}:{path}"
 
     def _reset_app(self):
+        self._get_id_by_root_path_map.clear_cache(self)
+        self._get_id_for_path.clear_cache(self)
         self._reset_app_cache_marker.clear_cache(self)
 
     @tools.ormcache()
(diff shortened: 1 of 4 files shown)


PicchiSeba:
The error seems unrelated to this PR

```
AttributeError: module 'marshmallow' has no attribute 'pprint'
```

fastapi/models/fastapi_endpoint.py
```diff
@@ -210,9 +210,24 @@ def _reset_app_cache_marker(self):
         to all the running instances.
         """
 
+    @api.model
+    def get_endpoint(self, root_path):
```
lmignon: The result should be cached to avoid a lot of useless process on each query???

lmignon (changes requested):
Than you for the proposal @PicchiSeba

`get_uid` , `get_app` and `get_endpoint` no more receive a root path. They receive a path... The parameter name should be renamed. (see also my previous comment)

fastapi/models/fastapi_endpoint.py
```diff
@@ -211,8 +211,24 @@ def _reset_app_cache_marker(self):
...
+            lambda fe: len(fe.root_path), reverse=True
+        )
+        endpoint = False
+        while endpoints_by_length:
+            candidate_endpoint = endpoints_by_length[0]
+            if path.startswith(candidate_endpoint.root_path):
+                endpoint = candidate_endpoint
+                break
+            endpoints_by_length -= candidate_endpoint
+        return endpoint
```
lmignon: This method should handle the path as a path :smirk:. The use of the `startswith`method doesn't mean that your endpoint path is a sub url of the given path...  e.g. /a/b is not prefix of /a/bc.

see https://github.com/PyTech-SRL/rest-framework/pull/1

lmignon (changes requested):
see https://github.com/PyTech-SRL/rest-framework/pull/1

lmignon:
once https://github.com/OCA/rest-framework/pull/527 is merged, a rebase will solve the CI

lmignon (approved):
Thank you for this great improvement @PicchiSeba 

LGTM (Code review, functionnal tests)

PicchiSeba:
> Thank you for this great improvement @PicchiSeba
> 
> LGTM (Code review, functionnal tests)

Thank you for your support and technical expertize!

lmignon:
ping @AnizR @sbidoul

fastapi/tests/test_fastapi_multislash_demo.py
```diff
@@ -0,0 +1,32 @@
...
+        )
+        cls.default_fastapi_authenticated_partner = cls.env["res.partner"].create(
+            {"name": "FastAPI Demo"}
+        )
+
+    def test_hello_world(self) -> None:
+        with self._create_test_client() as test_client:
+            response: Response = test_client.get("/demo/")
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertDictEqual(response.json(), {"Hello": "World"})
```
AnizR: I think that this test class should be re-worked:
You are not using your 'fastapi_endpoint_multislash_demo' correctly because you tried to put it on the 'default_fastapi_running_user' (defines the user not the endpoint).

AnizR (commented):
Besides the test class, the code seems good.
Thanks for your contribution!

fastapi/tests/test_fastapi_multislash_demo.py
PicchiSeba: Hello @AnizR, thanks for you review. I kinda forgot about testing properly with the Odoo test suite.

How would you test it?

I tried playing with the parameters of the `_create_test_client` method in  `common.py`, but I think it just creates a temporary client with the FastAPI app endpoints. I want to test that Odoo correctly mounts the FastAPI app with the specified url.

I also tried creating a test class inheriting from odoo.tests.HttpCase in order to use the `url_open` method but after activating the `fastapi.endpoint` it seems to be unreachable

simahawk: Have you checked that if the registry is in sync? You can call `_handle_registry_sync` directly if not.

PicchiSeba:
I accidentally squashed the new test into the merge commit, it should be good now

fastapi/tests/test_fastapi_multislash_demo.py
```diff
@@ -0,0 +1,32 @@
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/LGPL).
```
lmignon: IMO this test class is useless and should be removed. Your multi slash route is tested into the `test_multi_slash` method pf the existing test class.

lmignon (changes requested):
Thank you for the last changes. One little remaining comment.

PicchiSeba:
@lmignon I removed the useless test class

lmignon (approved):

AnizR (approved):

PR #490, opened 2025-01-16
[18.0][MIG] fastapi_auth_jwt: Migration to 18.0

dnplkndll:
Depends on:

- [ ] https://github.com/OCA/server-auth/pull/752
- [x] https://github.com/OCA/rest-framework/pull/471

PR #516, opened 2025-04-02
[MIG] fastapi_auth_jwt_demo: Migration to 18.0

kobros-tech:
- [ ] https://github.com/OCA/rest-framework/pull/490
- [ ] https://github.com/OCA/server-auth/pull/752
- [ ] https://github.com/OCA/rest-framework/pull/563


kobros-tech:
@sbidoul 
@lmignon 
@sebastienbeau 

can we review this?

kobros-tech:
@lmignon 

I would add this PR: https://github.com/OCA/rest-framework/pull/402
but the test for wrong URL or 401 is making error in the log, I know it is intended behaviour of yours but OCA bot does know understand that.

Is there a better way you have to run that test or shall I introduce some ideas?

fastapi_auth_jwt_demo/tests/test_fastapi_auth_jwt_demo.py
```diff
@@ -0,0 +1,146 @@
...
+        )
+        resp.raise_for_status()
+        whoami = resp.json()
+        self.assertEqual(whoami.get("name"), partner.name)
+        self.assertEqual(whoami.get("email"), partner.email)
+        self.assertEqual(whoami.get("uid"), self.env.ref("base.user_demo").id)
+        cookie = resp.cookies.get("demo_auth")
+        self.assertTrue(cookie)
+
+    def test_forbidden(self):
```
lmignon, line 84: @kobros-tech To solve the issue with the log you should decorate the method with a call to 'mute_logger' (`from odoo.tools import mute_logger`) to avoid the log of the stacktrace....
```suggestion
    @mute_logger('odoo.http')
    def test_forbidden(self):
```

lmignon (changes requested):

kobros-tech:
@lmignon 

now I see there is an issue with base_rest module js and import of js files.

I saw a couple of issues, but I don't know too much about that module and it's function?

kobros-tech:
@lmignon 

finally I fixed the issue in the other module and temporarily adding its commit here but last two commits should be deleted when we intend to merge.

PR #523, opened 2025-05-01
[17.0][MIG] rest_log: Migration to 17.0

halbtonjazz:
This is a migration PR

Task: 4619


lmignon:
could you also add #520

PR #524, opened 2025-05-05
[16.0][IMP]fastapi: obtain endpoint via common method

PicchiSeba:
Move endpoint matching into its own method to allow other modules to override it


lmignon (commented):
Than you for the proposal @PicchiSeba

`get_uid` , `get_app` and `get_endpoint` no more receive a root path. They receive a path...

PicchiSeba:
Hi @lmignon thank you for your review.

Do you want me to just change the names to `path` or to do something else?

EDIT: nvm, I just saw you other comment

lmignon:
closing since it will be merged with #515

Issue #391, opened 2023-11-06
[16.0]Fast API Endpoints Cache 


hassan510-cmd:
## Module
fast API

## Describe the bug
when I try to create a new app with such an endpoint as root_path into data dir and some routes,  it's required to update the module to be able to access 
this endpoint 

## To Reproduce
**Affected versions**:

1. create a new module depending on fastapi with some routes
2. create an endpoint in the data dir to be created during the installation 
3. try to access even the docs URL it'll not founded 
4. go to apps -> update your custom module
5. try to access it again, it works fine now

**Expected behavior**
the newly created endpoint should work properly without the need to update the module

lmignon:
@hassan510-cmd IMO this problem comes from the [endpoint_route_handler](https://github.com/OCA/web-api/tree/16.0/endpoint_route_handler) addon. @simahawk what do you thing about?

hassan510-cmd:
@lmignon i try to upgrade [endpoint_route_handler](https://github.com/OCA/web-api/tree/16.0/endpoint_route_handler) instead of upgrade my module, it work fine also

I think you are right

hassan510-cmd:
@lmignon @simahawk anything new about this issue ?

hassan510-cmd:
Dears,
after some dubbing, I have found that the first synced endpoint not working until you sync another one, and I tried to get an explanation for this behavior :
1- we have a table called "endpoint_route" This table contains new custom routes 
2- we have a sequence called "endpoint_route_version" that has 2 triggers before inserting and update
3- according to the incremental start with 1 and  INCREMENT BY 1 at first time the "last_version" col in  "endpoint_route_version" equal to 1
4- after syncing the first endpoint the last version does not change
[...]

PR #530, opened 2025-05-27
[16.0][FIX]fastapi: temporary fix to an incompatibility with the `endpoint_route_handler` module

PicchiSeba:
This PR is a temporary fix for https://github.com/OCA/rest-framework/issues/391 until the other module gets taken care of


lmignon (approved):
Thank you @PicchiSeba for the workaround. It's an ugly fixes indeed but it does the job and I'm not sure how to properly fix in into the base addon...

PR #531, opened 2025-06-04
[17.0][FIX] fastapi: Forwardport 16.0 pullrequest 486 - Avoid zombie threads

siemenv:
…threads


lmignon:
Thank you @siemenv for the forward port of this fixes. Nevertheless the authorship is not preserved. I'll try to align all the versions from 16 to 18 in the coming hours.

lmignon (changes requested):
commit history and authorship is not preserved

lmignon:
the forward port is included in #536 and preserve the commit history and authorship

PR #537, opened 2025-06-11
[16.0][IMP]fastapi: expose FastAPI docs only if field is set

PicchiSeba:
If we go into production right now, the FastAPI documentation endpoints (Swagger, Redoc, OpenAPI) are reachable by anyone.

There might be cases where this feature is not required or even prohibited.

With this PR we aim to hide these interfaces with a System Parameter


lmignon (changes requested):
I would prefer a configuration parameter on the endpoint in place of a global config parameter.

PicchiSeba:
While testing it I found a problem, not necessarily related to this implementation.

If I enable the endpoint with `expose_fastapi_docs` set to `True` the docs are reachable as expected.
Then if I change `expose_fastapi_docs` to `False the docs are still reachable.

I found that the `_prepare_fastapi_app_params` method is never triggered after synching the endpoint. Perhaps it gets computed only before mounting the FastAPI app?

lmignon:
> While testing it I found a problem, not necessarily related to this implementation.
> 
> If I enable the endpoint with `expose_fastapi_docs` set to `True` the docs are reachable as expected. Then if I change `expose_fastapi_docs` to `False the docs are still reachable.
> 
> I found that the `_prepare_fastapi_app_params` method is never triggered after synching the endpoint. Perhaps it gets computed only before mounting the FastAPI app?

It's because you must add the new field into the list returned by the method `_fastapi_app_fields`. This will instruct the addon that the app must be rebuild when you new field is modified.

fastapi/models/fastapi_endpoint.py
```diff
@@ -65,6 +65,7 @@ class FastapiEndpoint(models.Model):
         "unexpecteed disk space consumption.",
         default=True,
     )
+    expose_fastapi_docs = fields.Boolean("Expose FastAPI docs", default=True)
```
SirPyTech: **quibble**: We might use `expose_doc_urls` or something that does not explicitly mention fastapi since we are already in a FastAPI endpoint, and that would be more aligned to the URLs that it is hiding `docs_url`, `redoc_url`, `openapi_url`.

**suggestion**: Please mention the possibility to hide the docs in the documentation, for instance in https://github.com/OCA/rest-framework/blob/a0da317bf09ac7efad0c7e9055d6dbcbb6703012/fastapi/readme/USAGE.rst?plain=1#L169-L171.

fastapi/tests/test_fastapi.py
```diff
@@ -169,3 +175,24 @@ def test_no_commit_on_exception(self) -> None:
...
+                "app": "demo",
+                "demo_auth_method": "api_key",
+                "expose_fastapi_docs": False,
+            }
+        )
+        unexposed_endpoint._handle_registry_sync()
+        response = self.url_open(
+            self.endpoint_path(unexposed_endpoint, "docs"), timeout=20
+        )
+        self.assertTrue(400 <= response.status_code < 500)
```
SirPyTech: **suggestion**
```suggestion
        self.assertEqual(response.status_code, requests.codes.not_found)
```
because we actually want that URL to be "Not found", don't we?

fastapi/tests/test_fastapi.py
```diff
@@ -41,6 +42,11 @@ def _assert_expected_lang(self, accept_language, expected_lang):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.content, expected_lang)
 
+    def endpoint_path(self, endpoint, path):
+        if path[0] == "/":
+            path = path[1:]
```
SirPyTech, line 53: **question**: This is never used, why is it needed? Would you mind adding a comment with some explanation?

SirPyTech (changes requested):
Thanks for the PR!
Good work :clap:
I just have a few suggestions, mainly about documentation of the code, please have a look at the comments below.

fastapi/models/fastapi_endpoint.py
lmignon: +1

fastapi/tests/test_fastapi.py
SirPyTech: :+1:

fastapi/models/fastapi_endpoint.py
SirPyTech: Thanks!
Maybe you missed the documentation suggestion or maybe you don't want to do that, it's not blocking anyway :smile:

fastapi/tests/test_fastapi.py
SirPyTech: :+1:

SirPyTech (approved):

lmignon (approved):
@PicchiSeba Thank you once again. LGTM

fastapi/models/fastapi_endpoint.py
PicchiSeba: @SirPyTech I totally missed it, thank you for pointing it out!

HekkiMelody (approved):
Code review, LGTM

PR #436, opened 2024-06-04
[MIG][17.0] fastapi_auth_api_key: Migration to 17.0

sonhd91:
Working on a refactoring
* https://github.com/OCA/rest-framework/pull/418

PR #539, opened 2025-06-12
[17.0][MIG] add fastapi_auth_api_key

henrybackman:
Supercedes:
- https://github.com/OCA/rest-framework/pull/436


fastapi_auth_api_key/dependencies.py
```diff
@@ -0,0 +1,60 @@
...
+from odoo.addons.fastapi.dependencies import fastapi_endpoint, odoo_env
+from odoo.addons.fastapi.models.fastapi_endpoint import FastapiEndpoint
+
+from fastapi import Depends, status
+from fastapi.exceptions import HTTPException
+from fastapi.security import APIKeyHeader
+
+
+def authenticated_auth_api_key(
+    key: Annotated[str, Depends(APIKeyHeader(name="HTTP-API-KEY"))],
```
petrus-v, line 21: is it a real desired changed ? to avoid breaking consumer I would recomand to use same header as on previous version ie; https://github.com/OCA/rest-framework/blob/16.0/base_rest_auth_api_key/apispec/rest_method_security_plugin.py#L17


```suggestion
    key: Annotated[str, Depends(APIKeyHeader(name="API-KEY"))],
```

fastapi_auth_api_key/dependencies.py
```diff
@@ -0,0 +1,60 @@
...
+    try:
+        auth_api_key = admin_env["auth.api.key"]._retrieve_api_key(key)
+    except ValidationError as error:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED,
+            detail=error.args,
+            headers={"WWW-Authenticate": "HTTP-API-KEY"},
+        ) from error
+    # Ensure the api key is authorized for the current endpoint.
+    if auth_api_key not in endpoint.sudo().auth_api_key_group_id.auth_api_key_ids:
```
petrus-v, line 41: As far I understand we must set an `auth_api_key_group_id` to make the endpoint working, is that correct?

if yes I would set the field as required **or** support the case there is no `auth_api_key_group_id` defined:

```suggestion
    if endpoint.sudo().auth_api_key_group_id and auth_api_key not in endpoint.sudo().auth_api_key_group_id.auth_api_key_ids:
```

fastapi_auth_api_key/tests/test_fastapi_api_key_dependencies.py
```diff
@@ -0,0 +1,100 @@
...
+        self.assertEqual(error.exception.detail, "No HTTP-API-KEY provided")
+        # An exception is raised when no api key record is found
+        with self.assertRaises(HTTPException) as error:
+            authenticated_auth_api_key("404", self.demo_env, self.demo_endpoint)
+        self.assertEqual(error.exception.detail, ("The key 404 is not allowed",))
+        # TODO enable this when we know how to filter keys based
+        # on endpoint's api key group.
+        # An exception is raised when unauthorized api key record is found
+        # with self.assertRaises(HTTPException) as error:
+        # authenticated_auth_api_key("not_authorized", self.demo_env)
```
petrus-v, line 85: I'm not sure to understand the comment, maybe it's time on uncomment ?


```suggestion
        with self.assertRaises(HTTPException) as error:
            authenticated_auth_api_key("not_authorized", self.demo_env, self.demo_endpoint)
```

petrus-v:
I suppose this PR can be closed as superseded by https://github.com/OCA/rest-framework/pull/555

PR #541, opened 2025-06-16
[17.0] [FIX] fastapi: Running User Rule Domain

Tisho99:
Currently, the res.partner rule "fastapi: Running User Rule" filters records using the user_id field (Salesperson) instead of the user_ids field (Users).

image

Since the authenticated_partner_id context can be empty, the rule becomes too restrictive and may deny access to the running user, which defeats the purpose it was designed for.

T-8323


lmignon (approved):

etobella (approved):
/ocabot merge patch

PR #544, opened 2025-06-16
[17.0][ADD] fastapi_auth_apikey

Tisho99:
This module provides ``FastAPI`` ``Depends`` to allow authentication with Odoo's built in
[API Keys](https://www.odoo.com/documentation/master/developer/reference/external_api.html#api-keys)

T-8323


ValentinVinagre:
@simahawk @lmignon @etobella @sbidoul 
Let me explain the case a bit:
* Odoo APIKeys exist.
* You can't choose which FastAPI app uses each APIKey (or APIKey scope).
* This module allows you to choose APIKey authentication and specify a Scope based on the user selected in the app.
* This is a simple and maintainable use of Odoo APIKeys with FastAPI, using what Odoo has.

We've seen the module at https://github.com/OCA/server-auth/tree/17.0/auth_api_key . It provides an additional APIKey layer to the system and more maintenance, etc. However, it's more complex, and if you only want it for this FastAPI case, you'll have to maintain a much more complex module (I'm not saying it's incorrect, just that it's for more complex cases).

How would you view the use of this proposed concept?

lmignon:
@Tisho99  Thank you for the proposal. A first glance at the code suggests that this is very interesting. I'll try to take a closer look over the next few days.

ValentinVinagre:
> @Tisho99 Thank you for the proposal. A first glance at the code suggests that this is very interesting. I'll try to take a closer look over the next few days.

@lmignon , have you had time to review it? 😄

fastapi_auth_apikey/dependencies.py
```diff
@@ -0,0 +1,98 @@
...
+) -> Partner:
+    uid = env["res.users.apikeys"]._check_credentials(
+        scope=endpoint.apikey_auth_scope, key=api_key
+    )
+    partner = env["res.users"].sudo().browse(uid).exists().partner_id
+    if not partner:
+        raise HTTPException(
+            status_code=status.HTTP_401_UNAUTHORIZED, detail="Incorrect API Key"
+        )
+    if endpoint.user_id.id and endpoint.user_id.id != uid:
```
lmignon, line 37: @Tisho99 You link your endpoint to only one user? The `user_id` on the endpoint is the technical user running the process. The authentication is a way to authenticate a partner requesting the endpoint. I'm may be wrong when reading your code but this line will restrict the use of this endpoint to the partner linked to the user_id defined on the endpoint.

Tisho99: Hi @lmignon, sorry for the late response

Yes, my code works that way. Do you see any issue with that approach?

Since API keys in Odoo can be generated by any internal user, this effectively means that, without additional restrictions, any internal user could potentially gain access to the endpoint.

I’ll be thinking about a better approach

ValentinVinagre: @lmignon How do you see this approach? Do you think it would be better to use a checkbox to indicate which users are allowed?

yvaucher:
Is this a duplicate of #544 ?

ValentinVinagre:
> Is this a duplicate of #544 ?

This is the 544.

yvaucher:
My bad I meant #539

ValentinVinagre:
> My bad I meant #539

Owl-y shit, I hadn't seen the module. I think the idea is a little different since it's still done by scope, so it's not the idea of keeping two modules identical. I don't know what you think about this point of view; I understand that OCA would follow the other module. In my opinion, I'm not entirely convinced that you can see the API key logs. In this case, a scope is used to mask that point.
For our part, I have no problem canceling this PR.

yvaucher:
@ValentinVinagre As there was an existing module in v16 (but was never merged?)
It's arguable.

But to respect other work if @Tisho99 could base his work on it, that would be nice. (especially if it needs some migration scripts)

Otherwise I like the module name fastapi_auth_api_key better. With api and key separated as the camelcase APIKey translate to api_key in snake case.

ping @henrybackman @mmequignon

Tisho99:
Hi @yvaucher 

I also don't have problem closing this PR.

At glance, both modules are similar. This one uses Odoo core API Keys, and https://github.com/OCA/rest-framework/pull/539 uses OCA's

But I'd wait for more opinions

Thank you

ValentinVinagre:
> @ValentinVinagre As there was an existing module in v16 (but was never merged?) It's arguable.
> 
> But to respect other work if @Tisho99 could base his work on it, that would be nice. (especially if it needs some migration scripts)
> 
> Otherwise I like the module name fastapi_auth_api_key better. With api and key separated as the camelcase APIKey translate to api_key in snake case.
> 
> ping @henrybackman @mmequignon

Okay, so that's why we created this module, and the OCA module is based on the "auth_api_key_group" module, as @Tisho99 points out.
This module is based on the Odoo api-key concept.
Given the different approaches, I wouldn't join them. We can change the name to distinguish it. What do you think?
Fastapi_auth_odoo_api_key?

ValentinVinagre:
@Tisho99 change name to: Fastapi_auth_odoo_api_key or Fastapi_auth_scope_api_key

PR #545, opened 2025-06-17
[16.0][FIX] fastapi: Running User Rule Domain

lmignon:
backport of #541

### DIFF
--- a/fastapi/security/ir_rule+acl.xml
+++ b/fastapi/security/ir_rule+acl.xml
@@ -39,7 +39,7 @@
         <field name="model_id" ref="base.model_res_partner" />
         <field
             name="domain_force"
-        > ['|', ('user_id', '=', user.id), ('id', '=', authenticated_partner_id)]</field>
+        > ['|', ('user_ids', '=', user.id), ('id', '=', authenticated_partner_id)]</field>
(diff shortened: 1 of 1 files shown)


Tisho99 (approved):
LGTM

fastapi/security/ir_rule+acl.xml
```diff
@@ -39,7 +39,7 @@
         <field name="model_id" ref="base.model_res_partner" />
         <field
             name="domain_force"
-        > ['|', ('user_id', '=', user.id), ('id', '=', authenticated_partner_id)]</field>
+        > ['|', ('user_ids', '=', user.id), ('id', '=', authenticated_partner_id)]</field>
```
lmignon, line 42: @Tisho99 wouldn't it be more accurate to write `('user_ids', 'in', user.id)` ?

Tisho99: @lmignon 

Everything works with the Odoo ORM because they've designed their operators that way, but from what I know, the most commonly used nomenclature is the equals operator and the following: ('user_ids', 'in', [user.id]).

I think the most readable way would be to use the "like" operator, but I think it's also slower.

PR #547, opened 2025-06-25
[16.0][IMP] rest_log: Don't log retried error

lmignon:
If a retryable error occurs, it should only be logged if it will no more be retried by the retrying mechanism. This is required to avoid to log errors for requests that end successfully for the user thanks to the Odoo's retry mechanism.


sbejaoui (approved):
Very nice!

PR #551, opened 2025-07-28
[IMP] pydantic: moving from pydantic v1 to v2

petrus-v:
Smothly migrate webservice from pydantic<2.0 using pydantic odoo module to pydantic>2.0 getting the same behaviour retrieves odoo data from Odoo records.

Using this PR to migrate your pydantic models you have to do something this:

* before using `pydantic < "2.0"` your model looks like something like this:
```python

from pydantic import BaseModel

from odoo.addons.pydantic.utils import GenericOdooGetter


class PartnerModel(BaseModel):
    class Config:
        orm_mode = True
        getter_dict = GenericOdooGetter
    
    id: int
    name: str
```
* after using `pydantic > "2.0"` you should subclass `PydanticOdooBaseModel` which set the pydantic model config with the new `from_attributes=True` (aka old orm_mode). Your model will get same behaviour as the previous versions using the `GenericOdooGetter`:

```python
from odoo.addons.pydantic.utils import PydanticOdooBaseModel


class PartnerModel(PydanticOdooBaseModel):

    id: int
    name: str
```

### DIFF
--- a/pydantic/utils.py
+++ b/pydantic/utils.py
@@ -3,134 +3,82 @@
 
 from typing import Any
 
-from packaging.version import Version
-
 from odoo import fields, models
 
-from pydantic import __version__ as pydantic_version
+from pydantic import (
+    BaseModel,
+    ConfigDict,
+    ValidationInfo,
+    field_validator,
+    model_validator,
+)
 
-PYDANTIC_V2: bool = Version(pydantic_version) >= Version("2.0")
 
+class PydanticOdooBaseModel(BaseModel):
+    """Pydantic BaseModel for odoo record
 
-if PYDANTIC_V2:
-    from pydantic import (
-        BaseModel,
-        ConfigDict,
-        ValidationInfo,
-        field_validator,
-        model_validator,
+    This aims to help to serialize Odoo record
+    improving behavior like previous version:
+
+    * Avoid False value on non boolean fields
+    * Convert Datetime to Datetime timezone aware
+    * using int type on many2one return the foreign key id
+      (not the odoo record)
+    """
+
+    model_config = ConfigDict(
+        from_attributes=True,
     )
 
-    class PydanticOdooBaseModel(BaseModel):
-        """Pydantic BaseModel for odoo record
+    @classmethod
+    def model_validate(
+        cls,
+        obj: Any,
+        *,
+        context: Any | None = None,
+        **kwargs,
+    ):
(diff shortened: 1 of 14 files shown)


lmignon:
@petrus-v Thank you for the PR. IMO it's not a good idea to try to keep the compatibility with pydantic 1.  I'm not in favour of maintaining this support into OCA addon. Pydantic V1 is no more supported with fastapi, extendable_fastapi and a restriction is declared into the manifest file for these addons.

petrus-v:
@lmignon thanks for your feedback. I'm fine to drop the support to pydantic v1, at least I did it to write some unit test in the module to make sure serialisation behave in the same way. As far I understand your are not using the pydantic module in your v17.0 projects so we can drop the compatibility !

Let's do it !

petrus-v:
@lmignon et voilà : dropping support to pydantic v1 !

petrus-v:
@alexandregaldeano, @Koudja I would love to get your review on this one !

lmignon (approved):
Thank you @petrus-v

alexandregaldeano (approved):
LGTM

petrus-v:
@lmignon I've pushed a new commit that freeze graphql_server dependency to make the CI check pass to help resuming the merge process

pydantic/utils.py
```diff
@@ -5,64 +5,57 @@
...
+                    if issubclass(cls.__annotations__.get(info.field_name), int):
+                        # if field typing is an integer we return the .id
+                        # (not the odoo record)
+                        return value.id
+        return value
+
+    @model_validator(mode="before")
+    @classmethod
+    def odoo_model_validator(cls, data: Any, info: ValidationInfo) -> Any:
+        info.config["odoo_record"] = (
```
petrus-v: Just discover this probably create an issue in unittest testing twice the same endpoint

a workaround is to disable schema validaion on endpoint which can be disbaled like this `PYDANTIC_SKIP_VALIDATING_CORE_SCHEMAS=1` but would like to find a proper solution to store the record or change the ValidationInfo schema...

petrus-v: I'm now using context which avoid this issue and create an empty dict if not provide overloading `model_validate` classmehtod.

damdam-s (approved):
works perfectly on custom project !
thanks !

ThomasBinsfeld (approved):

PR #554, opened 2025-08-04
[16.0][ADD] fastapi_log

SirPyTech:
In the past months we have been proposing many improvements to https://github.com/OCA/rest-framework/pull/501 with https://github.com/akretion/rest-framework/pull/8.

From more than a month ago, the branch is outdated (https://github.com/akretion/rest-framework/pull/8#issuecomment-3018665362) and it is causing conflicts in https://github.com/akretion/rest-framework/pull/8.

Here we are proposing all those improvements to OCA directly.


SirPyTech:
@simahawk here I addressed your latest comments https://github.com/akretion/rest-framework/pull/8#discussion_r2204278496 and https://github.com/akretion/rest-framework/pull/8#issuecomment-3068521632, please have a look

api_log/models/api_log.py
```diff
@@ -0,0 +1,270 @@
...
+    _description = "Log for API"
+
+    collection_ref = fields.Reference(
+        selection="_selection_collection_ref",
+        index=True,
+    )
+    collection_model = fields.Char(
+        compute="_compute_collection",
+        store=True,
+        index=True,
```
simahawk, line 28: I'm not sure we need all these indexes. Looks like you search only using `collection_ref`. 
IMO you can drop the collection_ref index and search via model+id

api_log/models/api_log.py
```diff
@@ -0,0 +1,270 @@
...
+    response_time = fields.Float()
+
+    stack_trace = fields.Text()
+
+    # Derived fields
+    name = fields.Char(compute="_compute_name", store=True)
+    time = fields.Float(compute="_compute_time", store=True)
+    request_preview = fields.Text(compute="_compute_request_preview")
+    response_preview = fields.Text(compute="_compute_response_preview")
+    request_b64 = fields.Binary(
```
simahawk, line 59: what is the reason for these b64 fields?

api_log/models/api_log.py
```diff
@@ -0,0 +1,270 @@
...
+    )
+
+    @api.model
+    def _selection_collection_ref(self):
+        return []
+
+    @api.depends(
+        "collection_ref",
+    )
+    def _compute_collection(self):
```
simahawk, line 88: shouldn't be the other way around? You compute collection_ref based on model + id

api_log/models/api_log.py
```diff
@@ -0,0 +1,270 @@
...
+            if key in keys_to_hide:
+                headers_dict[key] = "<redacted>"
+        return headers_dict
+
+    @api.model
+    def _headers_to_dict(self, headers):
+        try:
+            headers_dict = {key: value for key, value in headers.items()}
+            return self._sanitize_headers_dict(headers_dict)
+        except AttributeError:
```
simahawk: when do you expect to have this kind of error?

api_log/models/api_log.py
```diff
@@ -0,0 +1,270 @@
...
+            "request_time": self._current_time(),
+        }
+        return log_request_values
+
+    @api.model
+    def log_request(self, request, override_log_values=None):
+        log_request_values = self._prepare_log_request(request)
+        if override_log_values:
+            log_request_values.update(override_log_values)
+        return self.sudo().create(log_request_values)
```
simahawk, line 154:
```suggestion
    def log_request(self, request, **kw):
        log_request_values = self._prepare_log_request(request)
        log_request_values.update(kw)
        return self.sudo().create(log_request_values)
```

api_log/models/api_log.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def _prepare_log_response(self, response):
+        return {
+            "response_status_code": response.status_code,
+            "response_headers": self._headers_to_dict(response.headers),
```
simahawk: One of the things done in rest_log was to inject a reference to the log entry in the response.
There we assumed the response was always JSON data and added a key.
Here we could inject a response header (eg: `API_LOG_ENTRY_URL`)

api_log/models/api_log_collection.py
```diff
@@ -0,0 +1,46 @@
...
+    log_ids = fields.One2many(
+        comodel_name="api.log",
+        compute="_compute_log_ids",
+        string="Logs",
+    )
+
+    def _compute_log_ids(self):
+        for collection in self:
+            collection.log_ids = self.env["api.log"].search(
+                [("collection_ref", "=", "%s,%s" % (collection._name, collection.id))]
```
simahawk: see my comment above on how to search for logs

api_log_mail/models/api_log.py
```diff
@@ -0,0 +1,41 @@
+# Copyright 2025 Simone Rubino - PyTech
+# License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+
+from odoo import api, models
+
+
+class FastapiLog(models.Model):
```
simahawk:
```suggestion
class ApiLog(models.Model):
```

simahawk (changes requested):

api_log/models/api_log.py
SirPyTech: 👍

api_log/models/api_log.py
SirPyTech: I thought they would be useful to store the binary value of the request/response, but now I see request/response body is already `Binary`.
Removed

api_log/models/api_log.py
SirPyTech: Sure, you can compute collection_ref based on model + id; or you can compute model + id from collection_ref.

I see no big difference between the two approaches, is there any specific reason to do it in the way you suggest?

api_log/models/api_log.py
SirPyTech: I'm suspicious about the `headers` argument, if it does not have the `items` method then an `AttributeError` would be raised.
But you're right, let's be optimistic and remove the exception handling :smile:

api_log/models/api_log.py
SirPyTech: I prefer to keep the log values in their own variable instead of using kwargs because this way more parameters can be added to the method to tweak its behavior.

I removed the `if` as suggested though.

api_log/models/api_log.py
SirPyTech: I have added the identifier (`api.log.id`) of the log record, that is enough to find the log record to whoever has access to them.

Please keep in mind that I agree with the author of https://github.com/OCA/rest-framework/pull/501 that wrote:
> This is not a migration of [rest_log](https://github.com/OCA/rest-framework/tree/16.0/rest_log), this module aims to be simpler, at least in the first iterations.

A more elaborated feature (like building the full URL of the log record) can be added later, after the first simple module is merged.

api_log/models/api_log_collection.py
SirPyTech: :+1:

api_log_mail/models/api_log.py
SirPyTech: Nice catch! Thanks

SirPyTech:
The error in the tests (https://github.com/OCA/rest-framework/actions/runs/17124124402/job/48571725570?pr=554#step:8:120) does not seem to be related to these changes
<details><summary>Stack</summary>
<pre>
2025-08-21 10:24:04,335 340 INFO odoo odoo.modules.loading: Loading module graphql_base (12/53) 
2025-08-21 10:24:04,475 340 CRITICAL odoo odoo.modules.module: Couldn't load module graphql_base 
2025-08-21 10:24:04,475 340 CRITICAL odoo odoo.modules.module: cannot import name 'HttpQueryError' from 'graphql_server' (/opt/odoo-venv/lib/python3.10/site-packages/graphql_server/__init__.py) 
2025-08-21 10:24:04,477 340 WARNING odoo odoo.modules.loading: Transient module states were reset 
2025-08-21 10:24:04,477 340 ERROR odoo odoo.modules.registry: Failed to load registry 
  File "/opt/odoo/odoo/modules/registry.py", line 87, in new
  File "/opt/odoo/odoo/modules/loading.py", line 493, in load_modules
    processed_modules += load_marked_modules(cr, graph,
  File "/opt/odoo/odoo/modules/loading.py", line 374, in load_marked_modules
    loaded, processed = load_module_graph(
  File "/opt/odoo/odoo/modules/loading.py", line 190, in load_module_graph
    load_openerp_module(package.name)
  File "/opt/odoo/odoo/modules/module.py", line 471, in load_openerp_module
    __import__('odoo.addons.' + module_name)
  File "/__w/rest-framework/rest-framework/graphql_base/__init__.py", line 4, in <module>
    from .controllers import GraphQLControllerMixin
  File "/__w/rest-framework/rest-framework/graphql_base/controllers/__init__.py", line 1, in <module>
    from .main import GraphQLControllerMixin
  File "/__w/rest-framework/rest-framework/graphql_base/controllers/main.py", line 6, in <module>
    from graphql_server import (
ImportError: cannot import name 'HttpQueryError' from 'graphql_server' (/opt/odoo-venv/lib/python3.10/site-packages/graphql_server/__init__.py)
2025-08-21 10:24:04,478 340 CRITICAL odoo odoo.service.server: Failed to initialize database `odoo`. 
  File "/opt/odoo/odoo/service/server.py", line 1333, in preload_registries
  File "<decorator-gen-16>", line 2, in new
  File "/opt/odoo/odoo/tools/func.py", line 87, in locked
    return func(inst, *args, **kwargs)
  File "/opt/odoo/odoo/modules/registry.py", line 87, in new
  File "/opt/odoo/odoo/modules/loading.py", line 493, in load_modules
    processed_modules += load_marked_modules(cr, graph,
  File "/opt/odoo/odoo/modules/loading.py", line 374, in load_marked_modules
    loaded, processed = load_module_graph(
  File "/opt/odoo/odoo/modules/loading.py", line 190, in load_module_graph
    load_openerp_module(package.name)
  File "/opt/odoo/odoo/modules/module.py", line 471, in load_openerp_module
    __import__('odoo.addons.' + module_name)
  File "/__w/rest-framework/rest-framework/graphql_base/__init__.py", line 4, in <module>
    from .controllers import GraphQLControllerMixin
  File "/__w/rest-framework/rest-framework/graphql_base/controllers/__init__.py", line 1, in <module>
    from .main import GraphQLControllerMixin
  File "/__w/rest-framework/rest-framework/graphql_base/controllers/main.py", line 6, in <module>
    from graphql_server import (
ImportError: cannot import name 'HttpQueryError' from 'graphql_server' (/opt/odoo-venv/lib/python3.10/site-packages/graphql_server/__init__.py)
</pre>
</details>

SirPyTech (commented):
Thanks for the review!
Please check the latest changes

api_log/models/api_log.py
paradoxxxzero: They were used to dowload the payload directly from the log form in case of a file upload/download.

api_log/models/api_log.py
```diff
@@ -0,0 +1,250 @@
...
+
+    @api.model
+    def log_request(self, request, override_log_values=None):
+        log_request_values = self._prepare_log_request(request)
+        log_request_values.update(override_log_values or {})
+        return self.sudo().create(log_request_values)
+
+    def _prepare_log_response(self, response):
+        headers_dict = self._headers_to_dict(response.headers)
+        headers_dict["API_LOG_ENTRY_ID"] = self.id
```
paradoxxxzero: This does not add the id to the headers of the response, only in the logs. Set it in the response instead.

api_log/views/api_log_views.xml
```diff
@@ -0,0 +1,116 @@
...
+                                name="request_preview"
+                                attrs="{'invisible': [('request_preview', '=', False)]}"
+                            />
+                            <field name="request_date" />
+                        </group>
+                        <group name="response">
+                            <field name="response_status_code" />
+                            <field name="response_headers_preview" />
+                            <field
+                                name="response_preview"
```
paradoxxxzero, line 42: There is no way to access the full response in the view anymore since the removal of the b64 field.
I think it should be kept one way or another.

fastapi_log/fastapi_dispatcher.py
```diff
@@ -0,0 +1,59 @@
...
+            try:
+                response = super().dispatch(endpoint, args)
+            except Exception as response_exc:
+                try:
+                    log and log.log_exception(response_exc)
+                except Exception as e:
+                    _logger.warning("Failed to log exception", exc_info=e)
+                else:
+                    # Be sure to commit/save the exception's log
+                    env.cr.commit()
```
paradoxxxzero: **This is very wrong.** You are committing a transaction of a failed request.

Consider this simple case :

```python
@router.post("/create_partner")
def create_partner(env: Annotated[api.Environment, Depends(optionally_authenticated_partner_env)]):
    partner = env["res.partner"].create({...})
    if is_something_wrong_with_new_partner(partner):
        raise UserError("Something is wrong with new partner")
    return "Ok"
```

If logging is enabled the wrong partner will be stored in database.

Please restore the separate cursor (https://github.com/OCA/rest-framework/pull/501/files#diff-67e6f9ad0c4d04c8de3dcd507c53fa5bbebf76b169c660d976f36c84c236ea68R38) or find another way to commit the exception.

paradoxxxzero (changes requested):
Good work thanks. Nothing blocking except the exception commit problem. Once it’s fixed I’ll close my old PR.

api_log/models/api_log.py
SirPyTech: Right, thanks for catching that!
I have added a quick assert in the tests too

fastapi_log/fastapi_dispatcher.py
SirPyTech: I verified this and you are correct.
I added the separate cursor in a commit with you as a co-author to fix this, could you please check?

api_log/models/api_log.py
SirPyTech: > They were used to dowload the payload directly from the log form in case of a file upload/download.

Thanks for letting us know, I added them back.

api_log/views/api_log_views.xml
SirPyTech: :+1: Answered to the same query in https://github.com/OCA/rest-framework/pull/554#discussion_r2269016610.

SirPyTech (commented):
Thanks for the review @paradoxxxzero, please check the latest changes

HekkiMelody (approved):

api_log/models/api_log.py
```diff
@@ -0,0 +1,269 @@
...
+        return log_request_values
+
+    @api.model
+    def log_request(self, request, override_log_values=None):
+        log_request_values = self._prepare_log_request(request)
+        log_request_values.update(override_log_values or {})
+        return self.sudo().create(log_request_values)
+
+    def _prepare_log_response(self, response):
+        response.headers["API_LOG_ENTRY_ID"] = self.id
```
paradoxxxzero: I think it could be better to use kebab-case here

api_log/models/api_log.py
```diff
@@ -0,0 +1,269 @@
...
+            "response_date": fields.Datetime.now(),
+            "response_time": self._current_time(),
+        }
+
+    def log_response(self, response):
+        log_response_values = self._prepare_log_response(response)
+        return self.sudo().write(log_response_values)
+
+    def _prepare_log_exception(self, exception):
+        values = {
```
paradoxxxzero, line 178: `API_LOG_ENTRY_ID` header is not set in case of exception

paradoxxxzero (approved):
Except for missing header in exception, LGTM

api_log/models/api_log.py
SirPyTech: Thanks, I added it, please check.

Note that it won't be returned in the response to the client but only stored in the log record, because in case of an exception the dispatcher simply propagates the exception raised by the endpoint.

api_log/models/api_log.py
SirPyTech: Right, it is the common convention for headers keys but I didn't know that, thanks!

api_log/models/api_log.py
```diff
@@ -0,0 +1,274 @@
...
+            "stack_trace": "".join(format_exception(exception)),
+            "response_headers": self._inject_log_entry(dict()),
```
paradoxxxzero: I think it is especially important to return the header in case of exception, luckily we can do this by setting exception headers attribute, see: https://github.com/OCA/rest-framework/blob/18.0/fastapi/fastapi_dispatcher.py#L46

```suggestion
        exception.headers = getattr(exception, "headers", {})
        values = {
            "stack_trace": "".join(format_exception(exception)),
            "response_headers": self._inject_log_entry(exception.headers),
        }
```
```

paradoxxxzero (changes requested):

api_log/models/api_log.py
SirPyTech: I agree, that's why I mentioned it specifically, but I didn't know a way to inject it; the solution you proposed is awesome and works great!
I included it, please check the new changes.

paradoxxxzero (approved):
LGTM

Also please rename this PR to something like: [16.0][ADD] fastapi_log

SirPyTech:
Rebased to check if this is affected by https://github.com/fastapi/fastapi/commit/51ad909ffe9f5b2d5c9315554e75e39a8a2d725c.
It still looks good :sunglasses:

paradoxxxzero:
@simahawk @lmignon Can we move forward on this PR? I would like to migrate these modules to 18.0

lmignon:
I don't use this addon. @simahawk as you made a review, is-it ok for you?

claudio-mano:
Hi, I'm currently trying this module, and I found a bug I'd want to report.

## Describe the bug

1. Endpoint raises an exception which has unwritable headers property (in my case oauthlib.oauth2.rfc6749.errors.MismatchingRedirectURIError)
1. fastapi_log tries to log exception
api_log._prepare_log_exception currently assigns exception.headers = getattr(exception, "headers", {})
1. Headers is a read-only property, so this triggers:
AttributeError: property headers of MismatchingRedirectURIError object has no setter

This is my traceback
oauthlib.oauth2.rfc6749.errors.MismatchingRedirectURIError: (invalid_request) Mismatching redirect URI. <oauthlib.Request SANITIZED>

During handling of the above exception, another exception occurred:

  File "/home/claudio/venv-enterprise/addons/extra/fastapi_log/fastapi_dispatcher.py", line 66, in dispatch
    log and log.log_exception(response_exc)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/claudio/venv-enterprise/addons/extra/api_log/models/api_log.py", line 202, in log_exception
    log_exception_values = self._prepare_log_exception(exception)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/claudio/venv-enterprise/addons/extra/fastapi_log/models/api_log.py", line 56, in _prepare_log_exception
    values = super()._prepare_log_exception(exception)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/claudio/venv-enterprise/addons/extra/api_log/models/api_log.py", line 175, in _prepare_log_exception
    exception.headers = getattr(exception, "headers", {})
    ^^^^^^^^^^^^^^^^^
AttributeError: property 'headers' of 'MismatchingRedirectURIError' object has no setter

The line of the error is this one
https://github.com/PyTech-SRL/rest-framework/blob/0fae517e7f93f2b2c65505bd9b91f2bc43240a25/api_log/models/api_log.py#L175

SirPyTech:
`pre-commit` will be fixed by https://github.com/OCA/rest-framework/pull/601.

SirPyTech:
Tests are failing too but they fail in https://github.com/OCA/rest-framework/pull/601 too so it seems not related to this PR.

SirPyTech:
@claudio-mano thanks for the detailed report in https://github.com/OCA/rest-framework/pull/554#issuecomment-4082942301, it helped a lot!
I added a commit to fix your use-case, could you please check if it works as expected now?

claudio-mano:
> @claudio-mano thanks for the detailed report in [#554 (comment)](https://github.com/OCA/rest-framework/pull/554#issuecomment-4082942301), it helped a lot! I added a commit to fix your use-case, could you please check if it works as expected now?

Hi @SirPyTech, thank you very much for the quick response, I confirm that now everything works fine

SirPyTech:
> > @claudio-mano thanks for the detailed report in [#554 (comment)](https://github.com/OCA/rest-framework/pull/554#issuecomment-4082942301), it helped a lot! I added a commit to fix your use-case, could you please check if it works as expected now?
> 
> Hi @SirPyTech, thank you very much for the quick response, I confirm that now everything works fine

Thanks for the feedback, would you mind dropping an approval in this PR? This would speed up the merge process.
If you don't know how to do that, look at the last step of https://odoo-community.org/resources/review:

> Once done, do not forget to leave a review on GitHub (accessible in the "Review changes" green button on the third tab - Files changed -) with:
> 
> -    Approve if everything worked as expected
> 
> -    Comment or Request changes with the description of the mismatch otherwise.
> 
> Anyone having a GitHub account can post a review.

claudio-mano (approved):
Tested my scenario described here: https://github.com/OCA/rest-framework/pull/554#issuecomment-4082942301

SirPyTech:
CI should be fixed after https://github.com/OCA/rest-framework/pull/601.

Rebased after merge of the linked PR.

lmignon:
@SirPyTech @paradoxxxzero Is this PR ready to merge?

SirPyTech:
> @SirPyTech @paradoxxxzero Is this PR ready to merge?

Thanks @lmignon for taking care of this PR!
Yes I fixed a small regression caught by the unit tests, caused by the latest changes.

I am probably biased because I'm the author of several commits, but in my opinion this could be merged :rocket:

PR #556, opened 2025-08-07
[FIX] rest_log: fix action window view mode

### LINKED ISSUE #553: Remaining tree reference
https://github.com/OCA/rest-framework/blob/25360135b4dc457f8df661954e9f161eeaf652bc/rest_log/views/rest_log_views.xml#L191

In the above line, `tree,form` is still referenced instead of the new `list,form`.

emadshaaban92:
Fixes #553


simahawk (approved):

simahawk:
Thanks!

PR #557, opened 2025-08-07
[FIX] base_rest_auth_api_key: fix method security plugin

emadshaaban92:


base_rest_auth_api_key/apispec/rest_method_security_plugin.py
```diff
@@ -18,7 +18,7 @@ def init_spec(self, spec):
         spec.components.security_scheme("api_key", api_key_scheme)
 
     def operation_helper(self, path=None, operations=None, **kwargs):
-        routing = kwargs.get("routing")
+        routing = kwargs.get("original_routing")
```
simahawk: we should probably use https://github.com/OCA/rest-framework/blob/18.0/base_rest/tools.py#L10

simahawk (commented):
thanks for your fix proposal!

base_rest_auth_api_key/apispec/rest_method_security_plugin.py
emadshaaban92: Thanks for your review, I rebased and made the change

simahawk (approved):
LGTM

lmignon (approved):

PR #563, opened 2025-09-03
[FIX] rest_base: correct file path importance

kobros-tech:
Suffexing js files with .esm.js is much better to correct style and syntax.


base_rest/__manifest__.py
```diff
@@ -20,8 +20,7 @@
     "assets": {
         "web.assets_frontend": [
             "base_rest/static/src/scss/base_rest.scss",
-            "base_rest/static/src/js/swagger_ui.js",
```
lmignon, line 23: No we should keep  js  files required for the swagger UI here. These are loaded into the dedicated view.

PR #566, opened 2025-09-25
[FIX][16.0] auth_partner: check if auth_partner.exists()

imbarbudiman:
browse() will return a pseudo-record.
Use browse().exists() to ensure the record exists.


paradoxxxzero (approved):

sebastienbeau (approved):
LGTM (code review)

PR #569, opened 2025-10-07
[19.0] [MIG] fastapi: Migration to 19.0

rven:


rven:
@lmignon @StefanRijnhart @pedrobaeza @sbidoul 
I notice that the CI pipelines are failing but this is related to the change in v19 of Odoo.
Odoo doesn't install demo data by default anymore when running Odoo.
```
+ unbuffer coverage run --include './*' --branch /opt/odoo-venv/bin/odoo -d odoo -i fastapi --test-enable --http-interface=127.0.0.1 --stop-after-init
```
You now have to specify the --with-demo argument yourself when demo data needs to be installed.

lmignon:
@rven IMO we should no more rely on demo data for tests. All tests have been adapted in Odoo, and we should do the same when migrating OCA modules.

rven:
@lmignon The power of the demo data is that a preset of data is loaded to do the validation when testing the functionality.
Like this module a lot of demo endpoints are created, which also can be tested in the UI.
If you change this in every module, it will be harder to test the functionality with already a preset of configured data.

pedrobaeza:
Yes, I was defending since a lot even before Odoo did the move that data used in tests should be populated in the test itself instead of using demo data. Now with this Odoo move, it makes it more mandatory, and IMO we shouldn't force to load demo data. The goal of the demo data is different than the tests, but to showcase the features of the module, or to provide basics for testing manually in runboat/runbot. One possible drawback is if runboat also don't load demo data, so if it's the case, then we should rethink it.

rven:
Odoo is still using demo data in their tests also, I don't see the problem to have this also enabled when running the tests.

lmignon:
@rven see https://github.com/DynAppsNV/rest-framework/pull/1 :smirk:

lmignon (approved):
Thank you for the migration. 
Code Review + Functional tests

abdullahoday710:
When is this expected to be merged ?

lmignon:
> When is this expected to be merged ?

@abdullahoday710 This PR depends on https://github.com/OCA/web-api/pull/117. It cannot be merged until the other one is merged. A review of these pending PRs will allow them to be merged.

dcorio:
Hi, thanks for the PR.
I had to add multilang: False to _get_routing_info() in order to use it on a multilang website.
It wasn't happening on v18...I think...

dannyadair (approved):
LGTM

len-foss (approved):
Tested, worked fine

PR #571, opened 2025-10-21
[19.0] Fixed compatibility issues with odoo19 for fastapi

abdullahoday710:
The latest odoo19 branch has introduced breaking changes for the module "fastapi"

This pull request makes the module installable on odoo19.

Key fixes : 

- Updated __manifest__.py to make the project installable and also bumped the base version to 19

- Odoo19 has changed the python class names for Partner and Users to ResPartner and ResUsers and fastapi module imported them directly, The import names have been fixed accordingly

- Odoo19 has changed the field name for users in "res.groups" from "users" to "user_ids" therefore res_groups.xml files has been patched with these changes

- Odoo19 removed the field "category_id" from the model "res.groups" for some reason, Therefore I have removed the field in res_groups.xml

- Odoo19 also changed the groups field name for the model "ir.ui.menu" to group_ids, So I have fixed that in fastapi_menu.xml

- Odoo19 search view definition doesn't allow the tag <group> anymore so I have removed it, Keeping only the "App" filter.

After testing the commit, The expected result is that the "fastapi" module can be installed on an odoo19 database.

### DIFF
--- a/fastapi/__manifest__.py
+++ b/fastapi/__manifest__.py
@@ -5,7 +5,7 @@
     "name": "Odoo FastAPI",
     "summary": """
         Odoo FastAPI endpoint""",
-    "version": "18.0.1.3.0",
+    "version": "19.0.0.0.0",
     "license": "LGPL-3",
     "author": "ACSONE SA/NV,Odoo Community Association (OCA)",
     "maintainers": ["lmignon"],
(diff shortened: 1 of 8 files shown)


abdullahoday710:
Apparently someone else is working on this, So I will close that PR

PR #572, opened 2025-10-31
[18.0] [FIX] fastapi: support both starlette < and >= 0.48

ivantodorovich:
Recent merged changes in a5bae2ba changed `status.HTTP_422_UNPROCESSABLE_ENTITY` to `status.HTTP_422_UNPROCESSABLE_CONTENT` in order to supress a deprecation warning emitted by the `starlette >= 0.48` library.

However, older versions of `starlette` will simply break with this change.

```
Module 'starlette.status' has no attribute 'HTTP_422_UNPROCESSABLE_CONTENT'.
Did you mean: 'HTTP_422_UNPROCESSABLE_ENTITY'?
```

~For this reason, we bump the minimum required version of `fastapi` to 0.120.0, which consumes newer versions of `starlette` where this is not an issue.~

To be compatible with past and future versions, we simply use the integer value `422`, aligning also with the updated `fastapi` documentation:

See:
- https://github.com/fastapi/fastapi/pull/14077

### DIFF
--- a/fastapi/error_handlers.py
+++ b/fastapi/error_handlers.py
@@ -29,7 +29,10 @@ def convert_exception_to_status_body(exc: Exception) -> tuple[int, dict]:
         status_code = exc.status_code
         details = exc.detail
     elif isinstance(exc, RequestValidationError):
-        status_code = status.HTTP_422_UNPROCESSABLE_CONTENT
+        # Use integer status code to supress starlette >= 0.48 deprecation warning
+        # See: https://github.com/fastapi/fastapi/pull/14077
+        # status_code = status.HTTP_422_UNPROCESSABLE_CONTENT
+        status_code = 422
         details = jsonable_encoder(exc.errors())
     elif isinstance(exc, WebSocketRequestValidationError):
         status_code = status.WS_1008_POLICY_VIOLATION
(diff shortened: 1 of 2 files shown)


lmignon (changes requested):
@ivantodorovich IMO We should revert the change related to `status.HTTP_422_UNPROCESSABLE_ENTITY` to support the same range of starlette version as in fastapi  "starlette>=0.40.0,<0.50.0". The depreciation has been introduced in starlette 0.48....

ivantodorovich:
Hi @lmignon !

What I don't like about that is that we will hit this issue in the future, at some point. Today it's a deprecation warning, but soon will be an exception.

But I see your point wanting to support older versions, too.. so perhaps we could do the same `fastapi` did in their project:
- https://github.com/fastapi/fastapi/pull/14077

That is, replace the use of these constants with their simple integer value (e.g.: `422`).
Or, maybe more elegantly, use the correct one depending on the `starlette` package version.

I'll update this PR with a new proposal

ivantodorovich:
@lmignon when you have a moment, please check the new proposal 🙏🏻 
Thanks!

lmignon (approved):

lmignon:
Thank you @ivantodorovich. This fix is pragmatic and solve the issue without introducing additional constrain.

SilvioC2C (approved):

PR #573, opened 2025-11-03
[18.0] [IMP] fastapi: show message for AccessError | MissingError

ivantodorovich:
It's common to raise, for example, `MissingError` when a related record is not found in a CREATE/WRITE operation.

e.g.: let's say the endpoint allows to create partner, providing the `country_code` to reference a valid country. If there's no country record with such country_code, we want to raise a MissingError with a meaningful message for the api consumer (e.g.: "No country found with code: %s")


lmignon:
Thank you for the proposals @ivantodorovich.  I understand the need. The current approach is very defensive and aims to avoid revealing the structure of the expected message in any way. In absolute terms, I am not against this change, Nevertheless this is a subject on which I would like to gather the opinions of as many people as possible.

SilvioC2C (approved):

PR #578, opened 2025-12-15
[16.0] [ADD] fastapi_endpoint_context

paradoxxxzero:
This module provides an overridable request context for FastAPI endpoints.

It allows to customize the behavior of requests deep in the odoo call stack
according to the current request context without having to override routers
that may come from external modules.

This context contains by default the current endpoint app and id, respectively
accessible via `self.env.context.get('fastapi_endpoint_app')` and
`self.env.context.get('fastapi_endpoint_id')`.

odoo_env dependency is automatically overridden to include FastAPI endpoint context values.

You can override the `fastapi.endpoint`->`_get_app_context()` method to add custom values to the context for your applications.


sebastienbeau (approved):
LGTM (code review)

Kev-Roche (approved):
Functionnal review ok !

PR #589, opened 2026-02-10
[18.0][IMP] rest_log: add profiling

grindtildeath:
[[REF] rest_log: Add hooks to read configuration from system params](https://github.com/OCA/rest-framework/commit/6c32cf32d18a4e4d4192abc8c706db8cb3bababd)

This allows to use the same syntax than the existing one from
rest.log.active to define other config parameters.

[[IMP] rest_log: Allow to profile requests through rest services](https://github.com/OCA/rest-framework/commit/f32377812ae351a47e14c6972164b51549ceac55)

Leverage ir.profile machinery to track what happens when a user
calls some rest services.

Since profiling can be done only for a single user at a time the
system parameter rest.log.profiling.uid has to be defined with
the ID from res.users whose calls need to be profiled.

Moreover, system parameter rest.log.profiling.conf can be defined
using the syntax from rest.log.active to fine tune which requests
will trigger the profiling for the user defined by the previous
parameter.

To note, once a request is identified to start the profiling, all
incoming requests from the user will be profiled until the timeout
duration defined by rest.log.profiling.duration is reached.


rest_log/components/service.py
```diff
@@ -39,6 +39,16 @@ class BaseRESTService(AbstractComponent):
     _log_calls_in_db = False
 
     def dispatch(self, method_name, *args, params=None):
+        if self._server_debug_logging_active(method_name):
```
simahawk: I wonder if we really need an IF here. All in all we can control this with debug levels.

rest_log/components/service.py
```diff
@@ -220,7 +230,53 @@ def _db_logging_active(self, method_name):
...
+            self.env["rest.log"]._get_matching_conf_from_param(
+                "rest.log.profiling.conf", self._collection, self._usage, method_name
+            )
+            and self.env.uid == profiling_uid
+        )
+        if res:
+            profiling_duration = 10
+            try:
+                profiling_duration = int(
+                    self.env["ir.config_parameter"]
```
simahawk: Do we really need to take care of such params?
We can simply set `base.profiling_enabled_until` when we set the other params.
We'll reduce complexity here. In fact, we could simply check if profiling is enable by calling `ir.profile._enabled_until()`

simahawk (commented):
early review, to be tested.

rest_log/components/service.py
```diff
@@ -220,7 +230,53 @@ def _db_logging_active(self, method_name):
...
+                    .sudo()
+                    .get_param("rest.log.profiling.duration")
+                )
+            except ValueError:
+                _logger.warning(
+                    "System parameter rest.log.profiling.duration is not defined. Profiling will be active for the next 10 minutes"
+                )
+            profiling_enabled_until = fields.Datetime.now() + relativedelta(
```
simahawk: missing imports

simahawk:
I've reworked it and simplified a bit. How it works:

1. set `rest.log.profiling.uid` to enable the given user id(s) (CSV list of ids)
2. set `rest.log.profiling.conf` to enable the conf for a specific app/collection (eg: `shopfloor.app`)

`base.profiling_enabled_until` is needed only to view the speedscope.

UPDATED: 12/02

rest_log/components/service.py
```diff
@@ -220,7 +226,47 @@ def _db_logging_active(self, method_name):
...
+            return None
+        profiling_uid = 0
+        try:
+            profiling_uid = int(
+                self.env["ir.config_parameter"]
+                .sudo()
+                .get_param("rest.log.profiling.uid")
+            )
+        except ValueError as err:
+            _logger.warning(
```
TDu, line 257: Should return None after ?

mmequignon: If no valid uid is found, then profiling_uid is set to 0, which never matches self.env.uid.
Then nothing happens.

rest_log/components/service.py
```diff
@@ -41,6 +42,9 @@ class BaseRESTService(AbstractComponent):
     def dispatch(self, method_name, *args, params=None):
         if not self._db_logging_active(method_name):
             return super().dispatch(method_name, *args, params=params)
+        if self._start_profiling(method_name):
+            with self._profiling_get_profiler():
```
grindtildeath: I see a mandatory arg not set here

simahawk: :facepalm:

simahawk:
using it live since a while: works fine :)

simahawk (approved):
Live on prod since months 🟢

simahawk:
Unrelated build issue https://github.com/OCA/rest-framework/issues/637

PR #596, opened 2026-02-25
[18.0][FIX] fastapi: create test data instead of relying on demo data

juancarlosonate-tecnativa:


lmignon:
@juancarlosonate-tecnativa Why in 18.0?

juancarlosonate-tecnativa:
> @juancarlosonate-tecnativa Why in 18.0?

In 18, in a fresh database, simply installing Fastapi and running the module tests causes the tests to fail.

lmignon:
> > @juancarlosonate-tecnativa Why in 18.0?
> 
> In 18, in a fresh database, simply installing Fastapi and running the module tests causes the tests to fail.

I agree on Odoo 19.0 but not on Odoo 18.0 otherwise the CI should fail from the creation of the branh 18.0 :thinking:

juancarlosonate-tecnativa:
> > > @juancarlosonate-tecnativa Why in 18.0?
> > 
> > 
> > In 18, in a fresh database, simply installing Fastapi and running the module tests causes the tests to fail.
> 
> I agree on Odoo 19.0 but not on Odoo 18.0 otherwise the CI should fail from the creation of the branh 18.0 🤔

To be honest, I can’t determine the real issue. Aren’t you experiencing this error on your side as well?

Also, if you check the Runbot-baseonly for this PR, you can see that the demo data is NOT being loaded either.

I repeat, simply installing FastAPI in a clean database and running the tests causes them to fail.

juancarlosonate-tecnativa:
@lmignon I’ve found the real issue. endpoint_route_handler does not declare any module as a dependency, which propagates demo=False and affects the tests. I’m closing the PR to fix it at the source.

PR #607, opened 2026-04-08
[18.0][IMP] fastapi: parse db param in sessionless requests

LoisRForgeFlow:
It is useful for multi-database environments.

PR #608, opened 2026-04-10
[19.0][IMP] fastapi: parse db param in sessionless requests

LoisRForgeFlow:
It is useful for multi-database environments.

Fwport of https://github.com/OCA/rest-framework/pull/607

### DIFF
--- a/fastapi/__init__.py
+++ b/fastapi/__init__.py
@@ -1,3 +1,4 @@
 from . import models
 from . import fastapi_dispatcher
 from . import error_handlers
+from . import patch
--- a/fastapi/patch.py
+++ b/fastapi/patch.py
@@ -1,0 +1,31 @@
+# Copyright 2026 ForgeFlow S.L. (https://www.forgeflow.com)
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/LGPL).
+
+from urllib.parse import parse_qs, urlparse
+
+from odoo.http import Request, db_filter
+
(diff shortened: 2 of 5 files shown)


lmignon (approved):
Thank you so much for the forward port @LoisRForgeFlow !

ThiagoMForgeFlow (approved):

sbidoul:
I have some worries about this. I had not noticed this feature before.

Is it not problematic that merely installing the `fastapi` module changes the way Odoo determines the database globally? Is it not preferable to do this in a standalone module that would be independent of fastapi?

LoisRForgeFlow:
> I have some worries about this. I had not noticed this feature before.
> 
> Is it not problematic that merely installing the `fastapi` module changes the way Odoo determines the database globally? Is it not preferable to do this in a standalone module that would be independent of fastapi?

Hi @sbidoul 

Odoo vanilla accepts db query param, so this is trying to make the same behavior happen with fastapi. I haven't seen the need to patch it in any other place.

Without `db` param odoo redirects you to the database selector. Example in a repo without this patch: http://oca-ddmrp-18-0-0cd23739c209.runboat.odoo-community.org/web/
image

Whereas if you include the db param you get redirected to that specific database: http://oca-ddmrp-18-0-0cd23739c209.runboat.odoo-community.org/odoo?db=b2a480f47-259f-4bdd-9d4c-dcb5194260ed-baseonly

image

What do you think?

PR #613, opened 2026-05-15
[19.0] [MIG] base_rest: Migration to 19.0

NachoAlesLopez:


NachoAlesLopez:
Can't seem to be able to reproduce the issues that come up in this PR pipeline, as both the zombie Chrome processes and the front-end styling missing the "_variables.scss" loading. Locally testing, both with and without demo, do pass the corresponding tests. Is this an issue specific to the pipeline configuration? Can something be done on the module's code so as to test that the "api-docs" page does report the API specification correctly? Aside, that is, manually calling the controller methods, which I could do as a last resort.

PR #614, opened 2026-05-17
[19.0][FIX] base_rest_pydantic: remove duplicate installable key

### LINKED ISSUE #612: base_rest_pydantic 19.0: Manifest file invalid
## Module
base_rest_pydantic

## Describe the bug
__manifest__.py contains duplicate "installable" keys, one of them residing inside "external_dependencies".
https://github.com/OCA/rest-framework/blob/7d9cb8e7605d99559cbd028716c8c5f1a8e2b3e1/base_rest_pydantic/__manifest__.py#L17

## To Reproduce
Install

1. Parse/validate the __manifest__.py file using any parsing tool that expects standard Odoo manifest schema.

**Expected behavior**
Manifest should conform to schema

**Additional context**
Add any other context about the problem here. (e.g. OS, Python version, ...)

fsmw:
**Issue:** #612

**Problem:**
The `__manifest__.py` contained a duplicate `installable` key. One instance was incorrectly nested inside the `external_dependencies` dict, making the manifest invalid according to the standard Odoo schema.

**Fix:**
- Removed the nested `installable` key from `external_dependencies`
- Moved the remaining `installable` key to the root dict with proper double quotes
- Bumped version from 18.0.1.0.2 to 18.0.1.0.3

**Before:**
```python
"external_dependencies": {
    "python": [
        "pydantic>=2.0.0",
    ],
    "installable": False,
},
    "installable": False,
}
```

**After:**
```python
"external_dependencies": {
    "python": [
        "pydantic>=2.0.0",
    ],
},
"installable": False,
}
```

Fixes #612

### DIFF
--- a/base_rest_pydantic/__manifest__.py
+++ b/base_rest_pydantic/__manifest__.py
@@ -14,7 +14,6 @@
         "python": [
             "pydantic>=2.0.0",
         ],
-    'installable': False,
-},
-    'installable': False,
+    },
+    "installable": False,
 }


lmignon:
@fsmw thank you for the fix. Can you let the version untouched so I can merge your fix. In practice, the version should never be modified by hand. The merge process will take car of it if appropriate.

base_rest_pydantic/__manifest__.py
```diff
@@ -5,7 +5,7 @@
     "name": "Base Rest Datamodel",
     "summary": """
         Pydantic binding for base_rest""",
-    "version": "18.0.1.0.2",
+    "version": "18.0.1.0.3",
```
lmignon:
```suggestion
    "version": "18.0.1.0.2",
```

lmignon (changes requested):

fsmw:
Hi @lmignon, thanks for the review! I've removed the version bump as requested. The version stays at `18.0.1.0.2` and the PR now only contains the fix for the duplicate `installable` key. Ready for re-review.

fsmw:
## CI failure is pre-existing

The CI failure (`test with Odoo` and `test with OCB`) is caused by `oca_checklog_odoo` flagging a `StarletteDeprecationWarning` from `fastapi/testclient.py` as an error — **this is unrelated to the manifest fix in this PR**.

Evidence: the **18.0 branch** has the exact same CI failure (run [26630008184](https://github.com/OCA/rest-framework/actions/runs/26630008184)):

```
StarletteDeprecationWarning: Using `httpx` with `starlette.testclient` is deprecated; install `httpx2` instead.
```

All 24 fastapi tests pass (0 failed, 0 errors). The only "error" is this deprecation warning from the `starlette` dependency, which `oca_checklog_odoo` treats as a failure.

This PR only touches `base_rest_pydantic/__manifest__.py` (removes duplicate `installable` key, fixes quotes) — the module has `installable: False` so it is not even installed during testing.

This appears to be a known issue that will need a separate fix in the `fastapi` addon (updating the `starlette`/`httpx` dependency).

PR #619, opened 2026-06-17
[19.0][MIG] fastapi_auth_api_key: Migration to 19.0

agiuseppe28:
Migration of `fastapi_auth_api_key` to 19.0.

History-preserving migration (`git format-patch | git am` from 18.0). The `[MIG]` commit:
- bump version to `19.0.1.0.0`
- `Partner` -> `ResPartner` (Odoo 19 model-class rename) in the type hints of `dependencies.py`
- lazy translation in `dependencies.py`: `env._("...") % {...}` -> `env._("...", HTTP_API_KEY_HEADER=...)` (fixes `translation-not-lazy`)
- adapt the test to `server-auth` 19.0: `auth.api.key._retrieve_api_key` now quotes the key (`The key '404' is not allowed`)

Dependencies available on 19.0: `fastapi` (this repo) and `auth_api_key_group` (`server-auth`).
Tests pass locally (7/7) on a fresh 19.0 database with demo data.

This module was not in the checklist of #567 (it landed on 18.0 in #568 after that issue was generated).

cc @mmequignon (maintainer)

PR #620, opened 2026-06-17
[19.0][FIX] test-requirements: use httpx2 for starlette TestClient

agiuseppe28:
On 19.0, `test-requirements.txt` pins `httpx`, but starlette's `TestClient` in this stack imports `httpx2` (`import httpx2 as httpx`). Without `httpx2` installed, the TestClient falls back to `httpx` and emits a `StarletteDeprecationWarning`, which the OCA CI log check treats as an error — failing `test with Odoo` / `test with OCB` for `fastapi` and every module depending on it (e.g. #619).

The 18.0 branch already uses `httpx2` here; the change didn't carry over when 19.0 diverged. This restores it, unblocking the test CI for fastapi-dependent modules on 19.0.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,2 +1,2 @@
 odoo_test_helper
-httpx
+httpx2


agiuseppe28:
@glitchov could you please approve? This would unlock several migrations. Thanks!

rvalyi (approved):
indeed, already httpx2 in 18.0 but as the files is not part of any Odoo module it was not carried to the 19.0 branch https://github.com/OCA/rest-framework/blob/18.0/test-requirements.txt#L2

rvalyi:
trivial, fast tracking...

PR #622, opened 2026-06-23
[16.0][IMP] api_log_mail: add default mail template for exceptions

alessandropecchini99:

PR #627, opened 2026-07-06
[17.0][IMP] api_log_mail: add default mail template for exceptions

alessandropecchini99:
FW Port from this PR: https://github.com/OCA/rest-framework/pull/622

### DIFF
--- a/api_log_mail/data/mail_template.xml
+++ b/api_log_mail/data/mail_template.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+    ~ Copyright 2025 Simone Rubino - PyTech
+    ~ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+-->
+<odoo noupdate="1">
+    <record id="email_template_fastapi_error" model="mail.template">
+        <field name="name">FastAPI: Error</field>
+        <field name="model_id" ref="api_log.model_api_log" />
+        <field
+            name="email_from"
+        >{{object.collection_ref.user_id.email or object.env.ref('base.partner_root').email}}</field>
+        <field
+            name="email_to"
+        >{{object.collection_ref.company_id.scheduler_error_email_address}}</field>
+        <field name="subject">Odoo FastAPI Error</field>
+        <field name="auto_delete" eval="False" />
+        <field name="body_html" type="html">
+            <div>
+                <t t-set="endpoint_name" t-value="object.collection_ref.name" />
+                <t t-set="request_url" t-value="object.request_url" />
+                <t t-set="request_method" t-value="object.request_method" />
+                <t t-set="headers" t-value="object.request_headers" />
+                <t t-set="exception" t-value="object.response_body" />
+                <t t-set="exception_stack" t-value="object.stack_trace" />
+
+                <p>Error in FastAPI call</p>
+                <p>FastAPI endpoint: <t t-out="endpoint_name" /></p>
+                <p>Request: [<t t-out="request_method" />] <t t-out="request_url" /></p>
+                <p>
+                    Headers:
+                    <ul>
+                        <li t-foreach="headers.items()" t-as="hd">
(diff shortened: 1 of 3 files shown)


HekkiMelody (approved):
Code review, LGTM

api_log_mail/data/mail_template.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<!--
+    ~ Copyright 2025 Simone Rubino - PyTech
+    ~ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+-->
+<odoo noupdate="1">
+    <record id="email_template_fastapi_error" model="mail.template">
```
SirPyTech, line 7: **suggestion**: `api_log_mail` is a generic module so it should not mention `fastapi`.
Please either remove the `fastapi` mentions, or move this to the `fastapi_log_mail` module.
Note that the migration from `16.0` should work as expected: for instance, if you rename the XMLID you should call https://oca.github.io/openupgradelib/API.html#openupgradelib.openupgrade.rename_xmlids.

api_log_mail/data/mail_template.xml
```diff
@@ -0,0 +1,47 @@
...
+        <field name="name">FastAPI: Error</field>
+        <field name="model_id" ref="api_log.model_api_log" />
+        <field
+            name="email_from"
+        >{{object.collection_ref.user_id.email or object.env.ref('base.partner_root').email}}</field>
+        <field
+            name="email_to"
+        >{{object.collection_ref.company_id.scheduler_error_email_address}}</field>
```
SirPyTech, line 15: **issue**: The field `scheduler_error_email_address` does not exist, so rendering the template raises
```
[...]
UserError: Failed to render inline_template template: {{object.collection_ref.company_id.scheduler_error_email_address}}
Error details: AttributeError("'res.company' object has no attribute 'scheduler_error_email_address'") while evaluating
'object.collection_ref.company_id.scheduler_error_email_address'
```
It would be a nice addition to the module, or you can just pick the email of the admin user.

After you fix it, remember to add a test so that there won't be a regression in the future, as suggested in:
> As a general rule, a bug fix should come with a unittest which would fail without the fix itself. This is to assure that regression will not happen in the future.

(from https://github.com/OCA/odoo-community.org/blob/master/website/Contribution/CONTRIBUTING.rst#6tests)

api_log_mail/data/mail_template.xml
SirPyTech, line 7: **chore**: Please inform the user in readme/CONFIGURE.md that this template exists and should be properly configured/customized.

api_log_mail/data/mail_template.xml
```diff
@@ -0,0 +1,47 @@
...
+        <field name="subject">Odoo FastAPI Error</field>
+        <field name="auto_delete" eval="False" />
+        <field name="body_html" type="html">
```
SirPyTech, line 18: **praise**: This prints out a nicely formatted email 🤗 
Image

api_log_mail/data/mail_template.xml
```diff
@@ -0,0 +1,47 @@
...
+    ~ Copyright 2025 Simone Rubino - PyTech
+    ~ License AGPL-3.0 or later (https://www.gnu.org/licenses/agpl).
+-->
+<odoo noupdate="1">
+    <record id="email_template_fastapi_error" model="mail.template">
+        <field name="name">FastAPI: Error</field>
+        <field name="model_id" ref="api_log.model_api_log" />
+        <field
+            name="email_from"
+        >{{object.collection_ref.user_id.email or object.env.ref('base.partner_root').email}}</field>
```
SirPyTech, line 12: **issue**: The field `user_id` does not exist in a generic collection, it is only present when the collection is a `fastapi.endpoint` record.
Please remove it if you choose to keep the template in this module.

SirPyTech (changes requested):
Thanks for the PR!
It would be nice to apply the changes directly in `16.0` too, especially the fix for the **issue**s.

PR #633, opened 2026-07-24
[18.0][MIG] base rest auth jwt

pierre-halleux:


lmignon (commented):
At first read It looks good to me. Need to be tested into a real project.

PR #616, opened 2026-06-12
[19.0][MIG] api_log, api_log_mail, fastapi_log, fastapi_log_mail: Migration to 19.0

TheClaud99:
Migration of the API/FastAPI logging modules to 19.0. They are submitted
together because they form a single dependency chain
(`api_log` → `api_log_mail`, `api_log` → `fastapi_log` → `fastapi_log_mail`).

## Modules

- `api_log`
- `api_log_mail`
- `fastapi_log`
- `fastapi_log_mail`

## Migration changes

- Bumped manifest versions to `19.0.1.0.0`.
- `api_log`: renamed the `res.groups` `users` field to `user_ids`
  (the field was renamed in 19.0).
- `fastapi_log`:
  - replaced the removed `Registry.in_test_mode()` with a `test_log_cr`
    attribute check in the FastAPI dispatcher;
  - updated the `TestCursor` import (`odoo.sql_db` → `odoo.tests.test_cursor`).
- Regenerated `README.rst` and `static/description/index.html` for 19.0.

Related to #567

PR #636, opened 2026-09-04
[19.0][FIX] test-requirements: pin anyio<4.15 to unblock CI

TheClaud99:
## Problem

`checklog-odoo` (`OCA_ENABLE_CHECKLOG_ODOO: "1"`) treats this record as an
error, failing the job after the addon tests have completed:

```
WARNING odoo py.warnings: starlette/testclient.py:53: DeprecationWarning:
The anyio.abc.BlockingPortal alias is deprecated, use anyio.from_thread.BlockingPortal instead.
```

This causes a red CI on every PR targeting `19.0` (like my PR here: #616)
## Root cause

anyio 4.15.0, released after the last run on `19.0`, turned
`anyio.abc.BlockingPortal` into a deprecated alias. `_PortalFactoryType` in
`starlette/testclient.py` is evaluated at import time, so merely importing
`starlette.testclient` emits the warning.

Timeline, since the branches still show green:

- anyio 4.15.0 published to PyPI: Sep 2
- last run on `19.0`: Aug 11
- last run on `18.0`: Sep 1

This is tracked upstream in Kludex/starlette#3497, with a fix proposed in
Kludex/starlette#3506 (opened Sep 4, not yet merged).
starlette 1.6.0 (latest) still ships the old alias, so no released version
avoids this today.

## Fix

Pin `anyio<4.15` in `test-requirements.txt`. This only constrains the test
environment. It can be dropped once a starlette **release** including
Kludex/starlette#3506 is available, not merely once that PR is merged.

CI on this PR is the verification: same code as `19.0`, green with the pin.

## Scope

Problably this problem affects every PR on this repo, not just the one where I hit it (#616).
`16.0`, `17.0` and `18.0` have the same import and need the same change; I can
open the equivalent PRs on those branches on request.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,2 +1,5 @@
 odoo_test_helper
 httpx2
+# starlette.testclient still uses the anyio.abc.BlockingPortal alias, which
+# anyio 4.15 deprecates: the resulting DeprecationWarning fails checklog-odoo.
+anyio<4.15


TheClaud99:
Upstream status: the fix was merged in Kludex/starlette#3498, which closed
Kludex/starlette#3497. It is not in a release yet: starlette on PyPI is still
1.6.0 (Aug 8), which ships the deprecated alias.

So the pin is still needed today, but it now has a clear expiry date. I'll
close this PR as soon as a starlette release including #3498 is out, or open
the revert if this has been merged by then.

simahawk (approved):
LGTM.

@lmignon ping

test-requirements.txt
```diff
@@ -1,2 +1,5 @@
 odoo_test_helper
 httpx2
+# starlette.testclient still uses the anyio.abc.BlockingPortal alias, which
+# anyio 4.15 deprecates: the resulting DeprecationWarning fails checklog-odoo.
+anyio<4.15
```
lmignon, line 5:
```suggestion
anyio!=4.15
```

lmignon (changes requested):
If it's the only affected version, i would prefer to only exclude one specific version.

test-requirements.txt
lmignon: Nevertheless thank you for the fix :smirk:

TheClaud99: Good point, but I think the problem isn't a regression in one release because the alias deprecation is intentional.
Version 4.15.1 was released on Sep 5 and behaves the same way (tested the warning in my local environment), so `<4.15` is the narrowest thing that works

lmignon: @TheClaud99  Let me try something to simply ignore this warning...

lmignon:
@TheClaud99 @simahawk see https://github.com/OCA/rest-framework/pull/639 to avoid to pin and lock to a specific version of anyio when it's not required by the api...

TheClaud99:
Closing in favour of #639, now merged. Thanks both.

PR #639, opened 2026-09-07
[19.0][FIX] CI: Ignore DeprecationWarning issued by anyio>=4.15

lmignon:
An alternative to #636


simahawk (commented):
LGTM. Yet, maybe we can have a comment on top of it to keep a reference without digging in logs?

lmignon:
> LGTM. Yet, maybe we can have a comment on top of it to keep a reference without digging in logs?

reference added.

TheClaud99 (approved):
LGTM

